PR #135, opened 2022-07-30
feat(kv): added aws dynamodb to kv capability

### LINKED ISSUE #103: create AWS S3 capability

Mossaka:
This PR adds AWS DynamoDB as the first AWS resource to KV capability. It uses aws-sdk-for-rust. Authentication is done by reading environment vars, accessKey and secretsKey. I've also added the necessary env vars to gh action secrets. This is not exactly what the issue is, but it is close #103. I used DynamoDB instead of S3 because it fits the KV model better.


danbugs (approved):
DynamoDB is a great example of a capability that could be implemented really well for our upcoming transactional WIT (`sql.wit`, perhaps), let's keep that in mind!

Other than that~ LGTM!

Issue #126, opened 2022-07-26
Add `slight-` as a prefix to all crates name


Mossaka:
This makes crates to be easily recognized as part of the slight implementation, and makes it easier to print trace logs:

`RUST_LOG=slight=trace` will enable log for all crates with prefix "slight"

PR #151, opened 2022-08-10
feat: added prefix slight to all crates

Mossaka:
We made a mistake earlier that didn't have prefix of "slight" to crate names. This causes Github to mistakenly think that a lot of packages are depending on slight crates. There is another benefit to the prefix: if we do RUST_LOG=slight=trace, it will show all tracing messages from all crates with the same prefix name. This will close https://github.com/deislabs/spiderlightning/issues/126

Signed-off-by: Jiaxiao Zhou [jiazho@microsoft.com](mailto:jiazho@microsoft.com)


danbugs (approved):
LGTM ~ funny, almost the same number of additions as deletions in this PR 😆

PR #153, opened 2022-08-11
implements az-app-config

### LINKED ISSUE #89: create Azure App Configuration implementation for config capability
As soon as #85 gets merged, our `config` capability will be capable of dynamically dispatching to two implementations of the interface:
  - `usersecrets`: quick and easy configs to set up development configs/secrets that encrypts values directly in your `slightfile.toml`, and
  - `envvars`: just uses the host environment's env vars.

In addition to these, we'd like to be able to dynamically dispatch to an implementation powered by [Azure App Configuration](https://docs.microsoft.com/en-us/azure/azure-app-configuration/overview).

**Additional context**
n/a

danbugs:
closes #89 

after this, I'll make a PR moving all of our configs to use az app config


crates/kv/Cargo.toml
```diff
@@ -15,7 +15,7 @@ slight-events-api = { path = "../events-api" }
 slight-runtime = { path = "../runtime" }
 slight-runtime-configs = { path = "../runtime-configs" }
 anyhow = "1"
-uuid = "1"
+uuid = { version = "1.1.2", features = ["v4"] }
```
Mossaka, line 18: This seems to be a very strict version. Was the feature "v4" require it to be equal or above "1.1.2"?

crates/runtime-configs/Cargo.toml
```diff
@@ -15,11 +15,13 @@ slight-runtime = { path = "../runtime" }
 argon2rs = "0.2.5"
 rand = "0.8.5"
 crossbeam-channel = "0.5.5"
-uuid = "1"
+uuid = { version = "1.1.2", features = ["v4"] }
 spiderlightning = { path = "../.." }
 slight-events-api = { path = "../events-api" }
 toml = "0.5.9"
 short-crypt = "1"
+azure-app-configuration = { git = "https://github.com/danbugs/azure-app-configuration.git", rev = "3c6c0886840847e98e66e9189f540d1db97fe316" }
```
Mossaka, line 23: Nice! Will you contribute this to the upstream `azure-sdk-for-rust`?

examples/configs-demo/usersecrets_slightfile.toml
```diff
@@ -10,7 +10,7 @@ value = "g88GVRuP1o2aqE6L1Mg"
 
 [[secret_settings]]
 name = "THIS_IS_ANOTHER_TEST_CONFIG_1061248"
-value = "g88GVRuP1o2aqE6L1Mg"
+value = "AJo6g9N6lABIM9HJb1A"
```
Mossaka, line 13: Why is this value kept changing?

crates/runtime-configs/src/implementors/azapp.rs
```diff
@@ -0,0 +1,55 @@
...
+
+#[cfg(test)]
+mod unittests {
+    use anyhow::Result;
+
+    use super::AzApp;
+
+    #[test]
+    fn set_then_get_test() -> Result<()> {
+        AzApp::set("key", "value".as_bytes())?;
```
Mossaka, line 51: I don't think you can run `AzApp::set` as unittest because it require connection to external provider. Unit test is supposed to be self-contained. How about port this to our integration test where you have access to environment variables?

Mossaka (commented):
Looks great to me. I wish there could be some examples showing how to use azapp to configure both the applications and the slight configuration.

.github/workflows/ci.yaml
```diff
@@ -41,3 +41,7 @@ jobs:
         run: make build-c
       - name: "Run Unit Tests"
         run: make test
+        env:
```
Mossaka, line 44: As the comment above, I don't think it's a good idea to run unit tests that read environment variables from the host machine. I will give another reason why this is not ideal.

Imagine if someone cloned our repo, and suddenly had no Internet access. At least they can do local development and run unittests to test their code. If you add external provider connections, then there is no way unittests will pass if there is no Internet connection.

crates/runtime-configs/src/implementors/azapp.rs
danbugs: I've moved this to integration tests ✅

.github/workflows/ci.yaml
danbugs: I've moved this to integration tests ✅

examples/configs-demo/usersecrets_slightfile.toml
danbugs: It's just from running the demo again.

crates/runtime-configs/Cargo.toml
danbugs: This isn't part of `azure-sdk-for-rust` ~ might contribute it to its' original repo, though

crates/kv/Cargo.toml
danbugs: Actually, I think we have a bunch of these occurrences~ I'll make an issue to fix our dependencies all around.

danbugs: done #160 ✅

tests/configs-test/usersecrets_slightfile.toml
```diff
@@ -0,0 +1,20 @@
...
+[[secret_settings]]
+name = "THIS_IS_ANOTHER_TEST_CONFIG_1061408"
+value = "m7N0Muoz0ICs7Owemrg"
+
+[[secret_settings]]
+name = "THIS_IS_ANOTHER_TEST_CONFIG_1061296"
+value = "g88GVRuP1o2aqE6L1Mg"
+
+[[secret_settings]]
+name = "THIS_IS_ANOTHER_TEST_CONFIG_1061248"
```
Mossaka, line 12: I've noticed that each run of config adds a new secret_settings to the slightfile. How can we avoid this?

Mossaka (approved):
LTGM, thanks for moving the unittests to integration tests. I've added a question.

PR #154, opened 2022-08-11
feat: Removed check and unittests from az pipeline

Mossaka:
I've noticed that, in avg, it takes 17 mins to execute `make improve` on windows machines. Considering that we are running both `make improve` and unittests on GH action, there is no need to run them again in our integration pipeline. This PR removes them, and hopefully will make our integration pipeline faster to finish. 

### DIFF
--- a/build/azure-pipeline/workflow-pr.yml
+++ b/build/azure-pipeline/workflow-pr.yml
@@ -132,22 +132,6 @@ stages:
           #
           # Build
           #
-          - script: make improve
-            displayName: Run rust-clippy and rustfmt
-            condition: and(succeeded(), eq(variables['target_os'], 'windows'))
-            env:
-              OPENSSL_DIR: C:\Program Files\OpenSSL-Win64\
-          
-          - script: make improve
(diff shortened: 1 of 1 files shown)


build/azure-pipeline/workflow-pr.yml
```diff
@@ -172,11 +156,8 @@ stages:
           #   displayName: Build C Examples
           
           #
-          # Running all tests
+          # Running integration tests
           #
-          - script: make test
```
danbugs, line 177: some tests might use environment variables, I don't think we should remove tests (e.g., #153)

danbugs (changes requested):
~one small request, otherwise, great idea!

build/azure-pipeline/workflow-pr.yml
Mossaka: we have github action dedicated to running unittests, so I believe it's okay to not run them again in our integration pipelines.

danbugs (approved):
LGTM

PR #156, opened 2022-08-11
changed every demo to use az app config rather than env vars

### LINKED ISSUE #155: change all examples to use `configs.azapp` for their secret store
n/a

**Additional context**
n/a

danbugs:
closes #155 

> Note: This PR is based off of PR #153, so, for ease of review, it should be checked after the other one is merged.


Mossaka (commented):
This mostly LGTM. Could you please add an example of user application that uses azure app configuration in the runtime? It could be as simple as fetching a key-value store's container name from the configuration provider.

Mossaka (approved):
LGTM

PR #169, opened 2022-08-21
refactor: removed unused config in Builder

Mossaka:
We are no longer using `config` in `Builder` anymore because #85 adds configuration capability to SpiderLightning that can take care of resource configurations. This PR removes this field from `Builder`. 


danbugs (approved):
LGTM!

devigned (approved):
lgtm!

Issue #170, opened 2022-08-22
address aws dynamo db test taking too long to run in some Windows builds


danbugs:
**Description of the bug**
see title

**To Reproduce**
n/a

**Additional context**
Here's a run where it happened: https://dev.azure.com/spiderlightning/slight/_build/results?buildId=94&view=logs&j=8c465dcb-f145-5732-cd9e-733cf3dc7b9c&t=45966b96-b6f1-5d78-ac3a-7f0e40cc01d5

danbugs:
This can only be completed after #230 is closed.

PR #171, opened 2022-08-23
temporarily removing windows from our matrix of pipelines

danbugs:
this addresses #170 


Mossaka (approved):
LGTM - please create a issue that brings windows pipelines back when we have more time to investigate.

PR #175, opened 2022-08-25
feat: new store and instance per http/events handler 

Mossaka:
## Problem

The current implementation of the HTTP capability uses a `Mutex` to guard shared states, like `Store` and `Instance`, in each HTTP request handler. The events capability shares a similar design. The use of locks is undesirable because it limits the throughput of the handlers and prevents http/events handlers to run concurrently. 

## Solution

Instead of using locks to guard shared states, we can pass in a state that implements the `Clone` trait. This is impossible for the current implementation because `wasmtime::Store` isn't clonable. The idea is to share a `Builder` that knows how to build a `wasmtime::Store` and `wasmtime::Instance` to http request handler or events handler. This PR does exactly this. 

## Crates dependencies
Since `Builder::build()` needs to know how to build `Store` and `Instance`, it needs to also know how to build each individual capability. In the current implementation, each capability implements `ResourceBuilder` trait and the runtime crate builds the resources and add them to wasmtime State in static time. Now, we are delaying adding states to the wasmtime Store at the runtime, because there isn't a Store anymore in build time. 

I've reversed the dependency in this PR from capability crates depending on runtime to runtime depending on all capabilities. This is reflected in `slight_runtime::ctx.rs` mod. Everything is working out nicely at the begining, but when I added http capability, the compiler started to fail, saying that there is a circular dependency, from http to runtime and then runtime to http. In order to build a store and instance, http crate will need a builder from runtime. But runtime also needs to import http to know how to build `slight_http::Http` resource. I used generic to resolve this problem, by making `Http` struct generic to the `Builder`.


crates/runtime/Cargo.toml
```diff
@@ -20,3 +20,13 @@ crossbeam-channel = "0.5"
 slight-events-api = { path = "../events-api" }
 slight-http-api = { path = "../http-api/" }
 tracing = { version = "0.1", features = ["log"] }
+slight-common = { path = "../common" }
+dyn-clone = "1"
+
+slight-kv = { path = "../kv" }
```
devigned, line 26: Having the runtime crate have a dependency on the capabilities seems backward. We should discuss if there is another way we should be structuring these crates. 

I don't think we need to address this in this PR, but rather, something we should discuss in September.

devigned (changes requested):
A couple superficial nits, and one topic I'd like us to discuss at a future date. 

The inversion of the runtime and generic builder are quite clever. I do wonder if there is a design approach where the runtime doesn't depend on the capabilities, or perhaps, if the runtime crate should be broken up.

In any case, it's great work and I'd like to see it pulled in after clean up the couple comment nits.

crates/runtime/Cargo.toml
Mossaka: I am thinking about making all capabilities generic to `T: Ctx` where `Ctx` is a context trait that defines the behaviors of a context. 

```rust
struct Kv<T: Ctx> {
  ...
}
```

crates/http/src/lib.rs
```diff
@@ -116,60 +111,57 @@ impl ServerInner {
 }
 
 /// Http capability
-#[derive(Default)]
-pub struct Http {
-    host_state: HttpState,
+// #[derive(Default)]
```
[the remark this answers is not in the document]
Mossaka: done

crates/runtime/Cargo.toml
Mossaka: Ah it was harder than I thought. I had some ideas about that requires more refactoring to the current design. 

Particularly, I want `HostState` to be cloneable.

devigned (approved):
lgtm!

PR #176, opened 2022-08-25
moved aws dynamo db from future resource examples to implemented resource examples in primer doc

danbugs:


devigned (approved):
lgtm

PR #177, opened 2022-08-25
contributing a full-fledged app for future demos

danbugs:
future apps can go under `app-demo` and we can elaborate onto the restaurant one too in the future.


examples/app-demo/restaurant-backend/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+wit_bindgen_rust::import!("../../../wit/http.wit");
+wit_error_rs::impl_error!(http::Error);
+
+wit_bindgen_rust::import!("../../../wit/mq.wit");
+wit_error_rs::impl_error!(mq::Error);
+
+fn main() -> Result<()> {
+    let router = http::Router::new()?;
+    let router_with_route = router
+        .post("/make_order", "handle_make_order")?
```
devigned, line 14: Perhaps, we can make these routes a bit more restful w/ a post to the orders resource.
```suggestion
        .post("/orders/", "handle_make_order")?
```

examples/app-demo/restaurant-backend/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+        .get("/get_next_order", "handle_get_next_order")?;
```
devigned, line 15: A `next` singleton resource hanging off of offers.
```suggestion
        .get("/orders/next", "handle_get_next_order")?;
```

examples/app-demo/restaurant-backend/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+fn handle_get_next_order(req: Request) -> Result<Response, Error> {
+    let mq = crate::mq::Mq::open("slight-restaurant").expect("failed to open message queue");
+    Ok(Response {
+        headers: Some(req.headers),
+        body: Some(mq.receive().expect("failed to get next order")),
+        status: 200,
+    })
+}
+
+#[register_handler]
```
devigned, line 42: Does this handler have a purpose? Is it for some illustrative purpose?

devigned (changes requested):
What do you think about adding a readme to describe the app, the routes, the behavior of the API, and how to run it?

Also added a couple suggestions about making the routes look a little more restful.

examples/app-demo/restaurant-backend/src/main.rs
danbugs: It is a requirement rn~ as per: https://github.com/deislabs/spiderlightning/blob/a2bd250eac68550d1df1d6b357d7d2aa1f89818e/examples/http-demo/src/main.rs#L75-L77

devigned (approved):
LGTM!

PR #179, opened 2022-08-26
chore: update azure blob version to 0.5

Mossaka:
Besides bumping up the version of azure blob storage to `0.5`, this PR also deletes one kv test in the integration test. It deletes the last bits that test to delete a key that doesn't exist in the container. Since AWS dynamodb and azure blob storage have different semantics over this operation, one silently passed and the other throwed an error. Before we decide to go either way, I will remove the test for now. 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -672,27 +672,26 @@ dependencies = [
 
 [[package]]
 name = "azure_core"
-version = "0.3.0"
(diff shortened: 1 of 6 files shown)


danbugs (approved):
LGTM~ just don't forget to run `make improve` prior to merging this.

PR #183, opened 2022-08-30
adding an implementor for pubsub that can be ran locally (i.e., Mosquitto)

danbugs:
This PR adds a new implementor for our `pubsub` capability: Mosquitto. This new implementor can be run locally, so we won't run into any issues of having examples that cannot be run due to lack of credits like we did for the `pubsub` Confluent Apache Kafka implementor.

Here's `slight` running the `pubsub-consumer-demo` and `pubsub-producer-demo` in tandem using the `mosquitto_slightfile.toml` configuration file:
image

> Note 1: To run this example locally, I installed mosquitto (`sudo apt install mosquitto`), and the mosquitto lib (`sudo apt install libmosquitto1`). In addition, when running the program for the first time, I ran into a linking issue and had to do the following:
> ```shell
> sudo ln -s `ldconfig -p | grep mosquitto | awk '{ print substr($4, 1, length($4)-2) }'` `ldconfig -p | grep mosquitto | awk '{ print substr($4, 1, length($4)-2) }'`
> ```
> (i.e., created a symbolic link to `libmosquitto.so.1` called `libmosquitto.so`)

> Note 2: While developing this, I noted that our `pubsub` interface is very much flawed, but this PR **DOES NOT** aim to fix the interface whatsoever, that work will come from another PR.

> Note 3: I'll be making a subsequent PR adding a "service-implementation101.md" doc documenting the process of making this implementor to show folks how they can do it themselves. I'll also be scheduling a meeting w/ people from the team to go through that document and re-implement the service — hoping to record that too and add it to the DeisLabs YouTube channel for the community at large.

Signed-off-by: Dan Chiarlone [dchiarlone@microsoft.com](mailto:dchiarlone@microsoft.com)


danbugs:
closes #184

devigned (changes requested):
My comments are mostly interface commentary, which is not the focus of this PR. However, I though it might be valuable for a future PR.

The capability code seems reasonable, and having a local implementation is definitely a big plus. I like that you have branched out into MQTT land as that's an area we haven't explored much but is valuable for tiny devices.

crates/pubsub/src/implementors/mosquitto.rs
```diff
@@ -0,0 +1,121 @@
...
+        msg_value: &[u8],
+        topic: &str,
+    ) -> Result<()> {
+        let formatted_message_with_key = &format!(
+            "{}-{}",
+            std::str::from_utf8(msg_key)?,
+            std::str::from_utf8(msg_value)?
+        );
+        // ^^^ arbitrarily formatting msg key and value like
+        // (as we have more implementors for pubsub, we should consider if we even
```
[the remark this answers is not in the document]
danbugs: ~agreed!

Also, I'm gonna keep this comment unresolved for future reference on the pubsub re-design 👍

crates/pubsub/src/implementors/mosquitto.rs
```diff
@@ -0,0 +1,121 @@
...
+// Sub
+impl MosquittoImplementor {
+    pub fn subscribe_to_topic(&self, topic: Vec<String>) -> Result<()> {
+        for t in topic {
+            self.subscriptions.lock().unwrap().push(t);
+        }
+        Ok(())
+    }
+
+    pub fn poll_for_message(&self, timeout_in_secs: u64) -> Result<String> {
```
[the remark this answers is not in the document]
danbugs: @devigned ~ do you mean we should have smt like a dictionary of vecs w/ messages under each topic (dict key)?

examples/configs-demo/usersecrets_slightfile.toml
```diff
@@ -10,7 +10,7 @@ value = "g88GVRuP1o2aqE6L1Mg"
 
 [[secret_settings]]
 name = "THIS_IS_ANOTHER_TEST_CONFIG_1061248"
-value = "jIybsBA7MIDM7B3VIjA"
+value = "_MSeeJP5ivT67vPUkwM"
```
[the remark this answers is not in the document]
danbugs: Yeah~ I've talked to Joe about this and I think we've decided not to take any actions on it yet. 

That's because this is **technically**  a feature — i.e., we all have different keys that were randomly generated by the usersecrets implementor of the configs capability, so the value will change as we each make PRs.

slight/src/commands/run.rs
```diff
@@ -166,7 +166,7 @@ fn build_store_instance(
                                 BasicState::new(resource_map.clone(), ss, toml_file_path),
                             )))?;
                     } else {
-                        bail!("the mq capability requires a secret store of some type (i.e., envvars, or usersecrets) specified in your config file so it knows where to grab the AZURE_SERVICE_BUS_NAMESPACE, AZURE_POLICY_NAME, and AZURE_POLICY_KEY from.")
+                        bail!("the pubsub capability requires a secret store of some type (i.e., envvars, or usersecrets) specified in your config file so it knows where to grab the AZURE_SERVICE_BUS_NAMESPACE, AZURE_POLICY_NAME, and AZURE_POLICY_KEY from.")
```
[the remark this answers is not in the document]
danbugs: oops, no~ good catch!

Makefile
```diff
@@ -35,6 +35,8 @@ install-deps:
 	sudo mkdir -p /opt/wasi-sdk
 	sudo mv wasi-sdk-15.0/* /opt/wasi-sdk/
 	sudo rm -rf wasi-sdk-*
+	sudo apt install libmosquitto1
+	sudo ln -s `ldconfig -p | grep mosquitto | awk '{ print substr($$4, 1, length($$4)-2) }'` `ldconfig -p | grep mosquitto | awk '{ print substr($$4, 1, length($$4)-2) }'`
```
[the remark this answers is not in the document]
danbugs: I've removed this in my latest push~ not necessary anymore ;)

slight/src/commands/run.rs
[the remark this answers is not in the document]
danbugs: done ✅

crates/pubsub/src/implementors/mosquitto.rs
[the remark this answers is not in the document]
devigned: I don't see a ton of use for key. If the key is used for deduplication, then perhaps. If the key is just more content, then why not add it to the message?

More broadly, what is the significance of the key? What is its purpose?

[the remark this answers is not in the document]
danbugs: No, I'm with you on this one — I don't think we need a key, and I definitely don't think this is something most pubsub services out there have.

crates/pubsub/src/implementors/mosquitto.rs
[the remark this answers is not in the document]
danbugs: resolving for checks sake, but noted ✅

crates/pubsub/src/implementors/mosquitto.rs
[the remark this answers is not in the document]
danbugs: resolving for checks sake, but noted ✅

crates/pubsub/src/implementors/mosquitto.rs
```diff
@@ -0,0 +1,157 @@
...
+        .unwrap();
+
+        Self {
+            mqtt: Arc::new(Mutex::new(mqtt)),
+            host,
+            port,
+            subscriptions: Arc::new(Mutex::new(Vec::new())),
+            subscriber: Arc::new(Mutex::new(None)),
+        }
+        // ^^^ arbitrarily chosing to create with 5 threads
```
Mossaka, line 69: what is this comment about? I didn't see the mentioned 5 threads

crates/pubsub/src/implementors/mosquitto.rs
```diff
@@ -0,0 +1,157 @@
...
+
+use anyhow::{Context, Result};
+use async_channel::Receiver;
+use futures::executor::block_on;
+use mosquitto_rs::{Client, Message, QoS};
+use slight_common::BasicState;
+
+#[derive(Clone)]
+pub struct MosquittoImplementor {
+    mqtt: Arc<Mutex<Client>>,
```
Mossaka, line 11: I would be careful about using `Arc<Mutex<...>>` because it has a performance overhead. In what scenerios did you see the necessity of `Arc<Mutex<...>>` on `mqtt`, `subscriptions` and `subscriber`?

crates/pubsub/src/implementors/mosquitto.rs
```diff
@@ -0,0 +1,157 @@
...
+    host: String,
+    port: i32,
+    subscriptions: Arc<Mutex<Vec<String>>>,
+    subscriber: Arc<Mutex<Option<Receiver<Message>>>>,
+}
+
+impl std::fmt::Debug for MosquittoImplementor {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "MosquittoImplementor")
```
Mossaka, line 20: Please add a TODO or FIXME here to remind us that we need to add a better debugging message to `fmt`. It is better to show the `mqtt` client struct.

crates/pubsub/src/implementors/mosquitto.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+}
+
+// Pub+Sub
+impl MosquittoImplementor {
+    pub fn new(slight_state: &BasicState) -> Self {
+        let mqtt = Client::with_auto_id().unwrap();
```
Mossaka, line 27: repalce `unwrap()` with a better error msg. I would actually make this function return a `anyhow::Result<Self>` and surface the error to the `pubsub::sub_open` to handle.

crates/pubsub/src/implementors/mosquitto.rs
```diff
@@ -0,0 +1,157 @@
...
+            )
+            .with_context(|| {
+                format!(
+                    "failed to get 'MOSQUITTO_PORT' secret using secret store type: {}",
+                    slight_state.secret_store
+                )
+            })
+            .unwrap(),
+        )
+        .unwrap()
```
Mossaka, line 58: There is a lot of `unwrap()`. My last comment about making this function returning a `anyhow::Result<>` could remove some of the unwraps.

crates/pubsub/src/lib.rs
```diff
@@ -206,6 +218,7 @@ impl SubInner {
 #[derive(Debug, Clone)]
 enum PubImplementor {
     ConfluentApacheKafka(PubConfluentApacheKafkaImplementor),
+    Local(MosquittoImplementor),
```
Mossaka, line 221: Why is it called Local instead of Mosquitto? "Local", to me, means that the implementor is purely relying on the underlying OS, like the OS File System, or OS Devices. I don't think Mosquitto is considered to be a OS component.

slight/src/commands/run.rs
```diff
@@ -166,7 +166,7 @@ fn build_store_instance(
                                 BasicState::new(resource_map.clone(), ss, toml_file_path),
                             )))?;
                     } else {
-                        bail!("the mq capability requires a secret store of some type (i.e., envvars, or usersecrets) specified in your config file so it knows where to grab the AZURE_SERVICE_BUS_NAMESPACE, AZURE_POLICY_NAME, and AZURE_POLICY_KEY from.")
+                        bail!("the pubsub capability requires a secret store of some type (i.e., envvars, or usersecrets) specified in your config file so it knows where to grab the CAK_SECURITY_PROTOCOL, CAK_SASL_MECHANISMS, CAK_SASL_USERNAME, CAK_SASL_PASSWORD, and CAK_GROUP_ID from.")
```
Mossaka, line 169: Great catch!

Makefile
```diff
@@ -96,8 +96,10 @@ run-rust:
...
 	RUST_LOG=$(LOG_LEVEL) $(SLIGHT) -c './examples/lockd-demo/slightfile.toml' run -m ./examples/lockd-demo/target/wasm32-wasi/release/lockd-demo.wasm
-	RUST_LOG=$(LOG_LEVEL) $(SLIGHT) -c './examples/pubsub-consumer-demo/slightfile.toml' run -m ./examples/pubsub-consumer-demo/target/wasm32-wasi/release/pubsub-consumer-demo.wasm &
-	RUST_LOG=$(LOG_LEVEL) $(SLIGHT) -c './examples/pubsub-producer-demo/slightfile.toml' run -m ./examples/pubsub-producer-demo/target/wasm32-wasi/release/pubsub-producer-demo.wasm
+	RUST_LOG=$(LOG_LEVEL) $(SLIGHT) -c './examples/pubsub-consumer-demo/mosquitto_slightfile.toml' run -m ./examples/pubsub-consumer-demo/target/wasm32-wasi/release/pubsub-consumer-demo.wasm &
+	RUST_LOG=$(LOG_LEVEL) $(SLIGHT) -c './examples/pubsub-producer-demo/mosquitto_slightfile.toml' run -m ./examples/pubsub-producer-demo/target/wasm32-wasi/release/pubsub-producer-demo.wasm
+	# RUST_LOG=$(LOG_LEVEL) $(SLIGHT) -c './examples/pubsub-consumer-demo/caf_slightfile.toml' run -m ./examples/pubsub-consumer-demo/target/wasm32-wasi/release/pubsub-consumer-demo.wasm &
```
Mossaka, line 101: remove commented out lines?

Mossaka (commented):
Great work on adding another implementation for the pubsub interface! 

I haven't tried it locally but I will do that shortly. The implementation looks great to me! I had a few comments on some Rust usages. I am looking forward to the next PR that improves the pain points you saw in the pubsub wit file!

crates/pubsub/src/lib.rs
danbugs: I'll change this.

crates/pubsub/src/implementors/mosquitto.rs
danbugs: If any of these things fail, I want to panic immediately and want to get the unwrap message.

Makefile
danbugs: I'll be bringing this example back — I'm downgrading my confluent account to the free one that doesn't use any credits, so I'm keeping it for now.

crates/pubsub/src/implementors/mosquitto.rs
danbugs: I'll make these `expect`s.

crates/pubsub/src/implementors/mosquitto.rs
danbugs: ~will do this.

crates/pubsub/src/implementors/mosquitto.rs
danbugs: That's right~ iirc, `subscriber` needed a `&mut self`, so I had to `Arc<Mutex<>>` it.

crates/pubsub/src/implementors/mosquitto.rs
danbugs: oops, that was from the old implementation~ good catch! I'll remove it (:

danbugs:
@Mossaka ~ When you run this locally, keep in mind you will need a local broker running (similar to how we need `etcd` running for the `lockd` examples). I just installed `mosquitto` and ran it with `mosquitto -p 1884`.

crates/pubsub/src/implementors/mosquitto.rs
danbugs: done ✅

crates/pubsub/src/implementors/mosquitto.rs
danbugs: done ✅

crates/pubsub/src/implementors/mosquitto.rs
danbugs: actually, I created an issue to improve the way we get configs for all crates: https://github.com/deislabs/spiderlightning/issues/191 ~ will be postponing the work 'till then, 'cause I think this same improvement could be applied in other places.

crates/pubsub/src/lib.rs
danbugs: done ✅

crates/pubsub/src/implementors/mosquitto.rs
Mossaka: That's very restricted, isn't it? I wonder why wit-bindgen doesn't give us a `&mut self_` for modifying the resources.

Mossaka (approved):
lgtm!

I made one minor comment on the `Arc<Mutex<..>>` thing but it's not related to this PR, but i'd like to hear your insights!

PR #192, opened 2022-09-07
pubsub refactor

danbugs:
# `pubsub.wit` refactor

## change 1

- Before:
```
send-message-to-topic: function(msg-key: payload, msg-value: payload, topic: string) -> expected<unit, error>
```

- After:
```
publish: function(message: payload, topic: string) -> expected<unit, error>
```

What changed?
- name (send-message-to-topic → publish), and
- eliminated the concept of keys (while keys are useful when strong ordering of messages is required, they are not present in most pubsub implementations I've seen out there).

## change 2

- Before:
```
subscribe-to-topic: function(topic: list<string>) -> expected<unit, error>
```

- After:
```
subscribe: function(topic: string) -> expected<unit, error>
```

What changed?
- name (subscribe-to-topic → subscribe), and
- now we subscribe to a single topic rather than a batch (the user interaction is now a bit less awkward it they just want to subscribe to a single topic).

## change 3

- Before:
```
poll-for-message: function(timeout-in-secs: u64) -> expected<message, error>
```

- After:
```
receive: function() -> expected<payload, error>
```

What changed?
- name (poll-for-message → receive), and
- now we no longer poll for messages for a specified amount of time, we just block on receive. Plus, as per change 1 (where we eliminated the key), we now return a `payload` (i.e., `Vec<u8>` in this case) instead of a `message` — the `message` type has been removed from `types.wit` too.

## epilogue

Plus, I've brought back our Kafka example by downgrading my account from a standard one (which used up all our free credits) to a free one — This refactor was tested, here's the output:
image


devigned (commented):
The design is much cleaner. 👍

The thing that jumps out at me is what we think about receive vs registering an event handler. @Mossaka do you have an opinion?

Mossaka:
> what we think about receive vs registering an event handler

It's a good question. It's contrasting two directions we can go with. One direction is to implement a `watch` function for each capability that makes sense to be observable. Another direction is to implement the PubSub capability for that component. 

For exmaple, we implemented `watch` for key-value filesystem. We can turn that using the second approach to an implementation of `PubSub` capability for the local filesystem.  

I would say that since `receive` is a blocking call, the pubsub implementation cannot handle concurrent events. However, the `watch` implementation, because it generates a Store and an Instance per event handler, can handle concurrent events. This is really a constraint from wit-bindgen, and I expect when `async` is stablized in upstream, we can totally choose one way to go.

Mossaka (approved):
lgtm!

PR #193, opened 2022-09-07
added config helper fxn

### LINKED ISSUE #191: improve the way we get secrets
~right now, getting secrets is rather cumbersome. We have to do something like:
```rs
        let port = String::from_utf8(
            slight_runtime_configs::get(
                &slight_state.secret_store,
                "MOSQUITTO_PORT",
                &slight_state.slightfile_path,
            )
            .with_context(|| {
                format!(
                    "failed to get 'MOSQUITTO_PORT' secret using secret store type: {}",
                    slight_state.secret_store
                )
            })
            .unwrap(),
        )
        .unwrap()
        .parse::<i32>()
        .unwrap();
```

... everytime we want a secret. 

For the Apache Kafka capability, I made a function like:
```rs
fn get_config(config_name: &str, state: &BasicState) -> Result<String> {
    let config = String::from_utf8(
        slight_runtime_configs::get(&state.secret_store, config_name, &state.slightfile_path)
            .with_context(|| {
                format!(
                    "failed to get '{}' secret using secret store type: {}",
                    config_name, state.secret_store
                )
            })?,
    )?;
    Ok(config)
}
```

... to avoid so much repetition, and I think a something like this needs to exist for all crates.

This issue is to track the work for that.

**Additional context**
n/a

danbugs:
closes #191, which was brought up by #183.

### DIFF
--- a/crates/kv/src/implementors/azblob.rs
+++ b/crates/kv/src/implementors/azblob.rs
@@ -3,6 +3,7 @@ use azure_storage::clients::StorageClient;
 use azure_storage_blobs::prelude::{AsContainerClient, ContainerClient};
 use futures::executor::block_on;
 use slight_common::BasicState;
+use slight_runtime_configs::get_from_state;
 
 use crate::providers::azure;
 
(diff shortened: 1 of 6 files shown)


Mossaka (approved):
lgtm!

PR #194, opened 2022-09-08
feat: Redis key-value capability implementation

### LINKED ISSUE #187: make redis implementor for kv capability
see title

**Additional context**
n/a

Mossaka:
This PR adds Redis implementation for the key-value capability. The configuration takes in a connection string like `redis://127.0.0.1:6379`. To be able to run the example, Redis server needs to be installed and started. 

image

This will close #187


Mossaka:
The redis integration test will need to add a bit more logic to ensure that `redis-server` is installed and running properly. I will defer that to another PR.

crates/kv/src/implementors/redis.rs
```diff
@@ -0,0 +1,62 @@
...
+            )
+            .with_context(|| {
+                format!(
+                    "failed to get 'REDIS_ADDRESS' secret using secret store type: {}",
+                    slight_state.secret_store
+                )
+            })
+            .unwrap(),
+        )
+        .unwrap();
```
danbugs, line 33: could you refactor this to use `get_from_state(.)` instead?

crates/kv/src/implementors/redis.rs
```diff
@@ -0,0 +1,62 @@
...
+        Self { client }
+    }
+
+    pub fn get(&self, key: &str) -> Result<Vec<u8>> {
+        let mut con = self.client.get_connection()?;
+        let val: Vec<u8> = con.get(key)?;
+        // Redis GET returns [:ok; nil] for non-existent keys
+        if val.is_empty() {
+            return Err(anyhow::anyhow!("key not found"));
+        }
```
danbugs, line 45: just checking, do you need to close the connection here?

crates/kv/src/implementors/redis.rs
```diff
@@ -0,0 +1,62 @@
...
+        let client = redis::Client::open(connection_string).unwrap();
+        Self { client }
+    }
+
+    pub fn get(&self, key: &str) -> Result<Vec<u8>> {
+        let mut con = self.client.get_connection()?;
+        let val: Vec<u8> = con.get(key)?;
+        // Redis GET returns [:ok; nil] for non-existent keys
+        if val.is_empty() {
+            return Err(anyhow::anyhow!("key not found"));
```
danbugs, line 44: use `bail!` instead of `return Err(anyhow::anyhow!(.))`

danbugs (changes requested):
LGTM~ just a couple of small changes requested!

crates/kv/src/implementors/redis.rs
Mossaka: sure

crates/kv/src/implementors/redis.rs
Mossaka: No, the connection will be dropped once the client is out of scope. https://docs.rs/redis/latest/redis/

danbugs (approved):
LGTM

PR #196, opened 2022-09-12
feat(test): added integration tests for kv redis implementaiton

Mossaka:
This PR adds the missing integration tests to the KV Redis resource. In fact, I found a bug in the previous redis implementation #187 that didn't use the `name` parameter in `open` at all. I used it and the key to form a compound key `name:key` which will differentiate one container name from another. 

The integration tests will only be run on UNIX OSs because Redis currently does not support Windows.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,11 +35,7 @@ install-deps:
 	sudo mkdir -p /opt/wasi-sdk
 	sudo mv wasi-sdk-15.0/* /opt/wasi-sdk/
 	sudo rm -rf wasi-sdk-*
-
-	curl -fsSL https://packages.redis.io/gpg | sudo gpg --dearmor -o /usr/share/keyrings/redis-archive-keyring.gpg
-	echo "deb [signed-by=/usr/share/keyrings/redis-archive-keyring.gpg] https://packages.redis.io/deb $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/redis.list
-	sudo apt-get update
-	sudo apt-get install redis
+	brew install redis
 
 .PHONY: install-deps-macos
 install-deps-macos:
--- a/Makefile
+++ b/Makefile
@@ -35,12 +35,6 @@ install-deps:
 	sudo mkdir -p /opt/wasi-sdk
 	sudo mv wasi-sdk-15.0/* /opt/wasi-sdk/
 	sudo rm -rf wasi-sdk-*
-	
-	curl -fsSL https://packages.redis.io/gpg | sudo gpg --dearmor -o /usr/share/keyrings/redis-archive-keyring.gpg
-	echo "deb [signed-by=/usr/share/keyrings/redis-archive-keyring.gpg] https://packages.redis.io/deb $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/redis.list
-	sudo apt-get update
-	sudo apt-get install redis
-	which redis-server
 
 .PHONY: install-deps-macos
 install-deps-macos:
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,11 @@ install-deps:
 	sudo mv wasi-sdk-15.0/* /opt/wasi-sdk/
 	sudo rm -rf wasi-sdk-*
 
+	curl -fsSL https://packages.redis.io/gpg | sudo gpg --dearmor -o /usr/share/keyrings/redis-archive-keyring.gpg
+	echo "deb [signed-by=/usr/share/keyrings/redis-archive-keyring.gpg] https://packages.redis.io/deb $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/redis.list
+	sudo apt-get update
+	sudo apt-get install redis
+
 .PHONY: install-deps-macos
 install-deps-macos:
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -35,6 +35,7 @@ install-deps:
 	sudo mkdir -p /opt/wasi-sdk
 	sudo mv wasi-sdk-15.0/* /opt/wasi-sdk/
 	sudo rm -rf wasi-sdk-*
+	brew install redis
```
danbugs, line 38: Is this a reliable installation of redis for Ubuntu? I know `brew` works just fine, just wondering if the tap is reliably updated.

crates/kv/src/implementors/redis.rs
```diff
@@ -13,18 +13,23 @@ use slight_runtime_configs::get_from_state;
...
+        Self {
+            client,
+            container_name,
+        }
     }
 
     pub fn get(&self, key: &str) -> Result<Vec<u8>> {
         let mut con = self.client.get_connection()?;
-        let val: Vec<u8> = con.get(key)?;
+        let val: Vec<u8> = con.get(format!("{}:{}", self.container_name, key))?;
```
danbugs, line 32: why?

crates/kv/src/implementors/redis.rs
```diff
@@ -34,14 +39,14 @@ impl RedisImplementor {
 
     pub fn set(&self, key: &str, value: &[u8]) -> Result<()> {
         let mut con = self.client.get_connection()?;
-        con.set(key, value)?;
+        con.set(format!("{}:{}", self.container_name, key), value)?;
```
danbugs, line 42: why?

crates/kv/src/implementors/redis.rs
```diff
@@ -34,14 +39,14 @@ impl RedisImplementor {
...
 
         Ok(())
     }
 
     pub fn delete(&self, key: &str) -> Result<()> {
         let mut con = self.client.get_connection()?;
-        con.del(key)?;
+        con.del(format!("{}:{}", self.container_name, key))?;
```
danbugs, line 49: why?

tests/integration.rs
```diff
@@ -91,6 +97,35 @@ mod integration_tests {
             run(SLIGHT, vec!["-c", file_config, "run", "-m", KV_TEST_MODULE]);
             Ok(())
         }
+
+        #[test]
+        #[cfg(unix)] // Redis is not available on Windows: https://redis.io/docs/getting-started/installation/install-redis-on-windows/
+        fn redis_test() -> Result<()> {
+            // make sure redis server is running
+            let port = get_random_port();
```
danbugs, line 105: Why use a random port? Does this ensure it is not being used?

danbugs (commented):
mostly LGTM, just some questions

Makefile
Mossaka: could you elaboate on your concerns? What problems would `brew` bring?

crates/kv/src/implementors/redis.rs
Mossaka: we want `folder_name:key` format to differentiate same keys from different named entities.

crates/kv/src/implementors/redis.rs
Mossaka: same as above

crates/kv/src/implementors/redis.rs
Mossaka: same as above

tests/integration.rs
Mossaka: yep

Makefile
danbugs: It could be serving an old version of redis because the brew tap isn't updated.

crates/kv/src/implementors/redis.rs
danbugs: Do we throw an error if the `folder_name` is equal?

Mossaka: I didn't understand your question. If folder_name is the same, then we are performing `get`, `set` and `delete` on that folder. I just don't want the following to happen:

```rust
let kv1 = Kv::open("container-a");
let kv2 = Kv::open("container-b");
assert!(kv1.get("key") == kv2.get("key"));
```

danbugs: I was just thinking of the use case where we do:
```rs
let kv1 = Kv::open("container-a");
let kv2 = Kv::open("container-a");
```

Makefile
danbugs: Yeah~ My concern was mostly that the linux tap and the macos one might have diff versions, but, if your experience has been ok, I'm good to use brew as a package manager for our Ubuntu tests too!

crates/kv/src/implementors/redis.rs
Mossaka: Ah, so in the above example, if `kv1` writes a key, then `kv2` will see it.

Makefile
Mossaka: I think it will be okay for now~

danbugs (approved):
LGTM

PR #197, opened 2022-09-12
chore!: Bump wasmtime to 0.39

Mossaka:
This PR does a few things:
1. bump wasmtime to `0.39.1`
2. bump wit-bindgen to tag `v0.2.0`
3. removed module linking
4. update wit syntax `function` -> `func`
5. update wasmtime and wit on examples
6. update docs

### DIFF
--- a/crates/common/Cargo.toml
+++ b/crates/common/Cargo.toml
@@ -14,5 +14,5 @@ doctest = false
 slight-events-api = { path = "../events-api" }
 slight-http-api = { path = "../http-api/" }
 as-any = "0.3"
-wasmtime = "0.35"
+wasmtime = "0.39.1"
 anyhow = "1"
--- a/crates/common/Cargo.toml
+++ b/crates/common/Cargo.toml
@@ -14,5 +14,5 @@ doctest = false
 slight-events-api = { path = "../events-api" }
 slight-http-api = { path = "../http-api/" }
 as-any = "0.3"
-wasmtime = "0.39.1"
+wasmtime = "0.39"
 anyhow = "1"
--- a/crates/http/Cargo.toml
+++ b/crates/http/Cargo.toml
@@ -8,9 +8,9 @@ authors = ["DeisLabs Engineering Team"]
 doctest = false
 
 [dependencies]
-wasmtime = "0.35"
+wasmtime = "0.39.1"
 anyhow = "1"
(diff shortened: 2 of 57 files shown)


devigned (approved):
lgtm!

crates/common/Cargo.toml
```diff
@@ -14,5 +14,5 @@ doctest = false
 slight-events-api = { path = "../events-api" }
 slight-http-api = { path = "../http-api/" }
 as-any = "0.3"
-wasmtime = "0.35"
+wasmtime = "0.39.1"
```
danbugs, line 17: Why use `0.39.1` instead of `0.39`? Across all other deps in the project, we've been sticking to only `major.minor` semver

crates/http/Cargo.toml
```diff
@@ -8,9 +8,9 @@ authors = ["DeisLabs Engineering Team"]
 doctest = false
 
 [dependencies]
-wasmtime = "0.35"
+wasmtime = "0.39.1"
 anyhow = "1"
-wit-bindgen-wasmtime = { git = "https://github.com/Mossaka/wit-bindgen", rev = "8252b0e39c7495f647ec0b0898721a7c641fc6c8" }
+wit-bindgen-wasmtime = { git = "https://github.com/Mossaka/wit-bindgen", rev = "f7734c4c5c306bd8f0c047bb4eba782c5cdbc416" }
```
danbugs, line 13: can we use a tag here too?

danbugs (commented):
mostly LGTM, just some questions

crates/http/Cargo.toml
Mossaka: I would rather keep this rev here since this is from my forked repo. This allows me to track which exact commit are we using in slight.

crates/common/Cargo.toml
Mossaka: good point!

crates/http/Cargo.toml
danbugs: Why are we using your forked repo? Will you be contributing your work upstream?

Mossaka: You can see what I added to wit-bindgen. https://github.com/bytecodealliance/wit-bindgen/compare/main...Mossaka:wit-bindgen:pub_fields

Probably not because this is not considered a good practice to make fields public. I'd convert this to an issue and figure out another way to solve this.

danbugs (approved):

PR #198, opened 2022-09-12
fix(http): allow http cap be used without the default handler

Mossaka:
The current http capability experience is not great, since people have to decalre a http_handler function in the wasm module, due to the generated `HttpHandler::new` function that looks for that function. This PR allows http capability be used without the need to declare http-handler function in wasm modules anymore.

See this file for the effect of this work `examples/http-demo/src/main.rs`

This PR and a[ PR on a customized wit-bindgen repo ](https://github.com/bytecodealliance/wit-bindgen/commit/6f80d65a483fe929a79e2661825d99842227b146)do a few things:
1. make `get_state` in generated code public (in wit-bindgen)
2. defined a wrapper for `HttpHandler` that can take any handler-name in `http-api` crate
3. `http` crate uses the `HttpHandler` to call http handler from `http-api` crate
4. remove unused `wasmtime` dependency from `http` crate

> Note: this PR builds on top of #197 and will wait to be merged in until #197 merges in.

### DIFF
--- a/.devcontainer/setup.sh
+++ b/.devcontainer/setup.sh
@@ -29,7 +29,7 @@ rustup component add clippy --toolchain nightly
 
 cargo install cargo-expand
 cargo install cargo-edit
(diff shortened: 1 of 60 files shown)


danbugs (approved):
LGTM

PR #201, opened 2022-09-14
fix(runtime): removed unused preopen that causes issues when running slight outside…

Mossaka:
… of the spiderlightning repo


danbugs (approved):
LGTM

PR #202, opened 2022-09-15
feat: change path of type `&str` to type `AsRef<Path>`

Mossaka:
In library design, it is more versatile to use `AsRef<Path>` to abstract any path types. Thie allow the caller to path `Path`, `PathBuf`, `&str` or `String`.


danbugs (approved):
LGTM

PR #206, opened 2022-09-20
fix(http): Removed block_on on http body parsing

Mossaka:
The comment that explained why there was a `block_on` on http body parsing was no longer applicable, since #175 removed the lock on store and instance, instead creates them for each http request. 


devigned (approved):
LGTM

Issue #208, opened 2022-09-27
Calling kv::set will freeze the HTTP handler


Mossaka:
**Description of the bug**
That is, using kv capability inside of a http handler will render it frozen. 

**To Reproduce**
```rust
#[register_handler]
fn handle_bar(request: Request) -> Result<Response, Error> {
    assert_eq!(request.method, Method::Put);
    if let Some(body) = request.body {
        let kv = crate::Kv::open("my-container").unwrap();
        kv.set("key", &body).unwrap();
    }
    Ok(Response {
        headers: Some(request.headers),
        body: None,
        status: 200,
    })
}
```

**Additional context**

Mossaka:
The cause of the issue: the implementation of HTTP service uses tokio runtime, while az blob storage and aws dynamodb use `futures::executor` runtime.

Mossaka:
Closed by #209

PR #209, opened 2022-09-27
feat(all): add async to http, kv, and runtime-configs

Mossaka:
This PR uses wasmtime async feature to remove the use of `block_on` in some crates, which is the root cause for #208 . There are some leftover work to be done. 
- [x] remove `block_on` from kv cap
- [x] remove `block_on` from runtime-configs cap
- [x] remove `block_on` from http cap
- [x] remove `block_on` from lockd cap
- [x] remove `block_on` from mq cap
- [x] remove `block_on` from pubsub
~[ ] remove `block_on` from events~

The events block_on is necessary becasue it is in a OS thread. I will create an issue just for it. 


danbugs (approved):
LGTM~ very easy to review!

PR #212, opened 2022-09-30
implemented add command

### LINKED ISSUE #35: [slight] impl add command to slight to download interfaces of a specific version
What's the user experience we want for developers making back-end services w/ Slight? That is, say that they have slight installed on their local machine or server... How do we want them do make use of our interfaces? Are we expecting them to just to go GitHub and copy-and-paste the `.wit` file?

I think it'd be interesting to, one day, give our slight CLI app the functionality of allowing the user to add capabilities to their project.

`$ slight add kv@v0.1.0`

After this, download the interface.

danbugs:
Here's how the interaction is looking:
image

The `add` command will pretty much just curl the wit file artifact (i.e., the interface, and its' dependencies) from a specific release tag and put it under a folder called `<capability_name>_<release_tag>`.

closes #35 

### DIFF
--- a/slight/src/commands/add.rs
+++ b/slight/src/commands/add.rs
@@ -1,0 +1,84 @@
+use std::{
+    fs::{create_dir_all, remove_dir_all, File},
+    io::{self, ErrorKind},
+};
+
+use anyhow::Result;
+
+const GITHUB_URL: &str = "https://github.com/deislabs/spiderlightning/releases/download";
+
+const KV_DOWNLOADS: [&str; 3] = ["kv", "types", "resources"];
+const CONFIGS_DOWNLOADS: [&str; 2] = ["configs", "types"];
+const EVENTS_DOWNLOADS: [&str; 4] = ["events", "event-handler", "types", "resources"];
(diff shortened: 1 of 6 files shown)


slight/src/commands/add.rs
```diff
@@ -0,0 +1,84 @@
...
+        // ^^^ fine to unwrap, we are guaranteed to have a '@' at this point.
+        (
+            &what_to_add[..find_at],
+            &what_to_add[find_at + 1..],
+            what_to_add.replace('@', "_"),
+        )
+    };
+
+    match what_to_add {
+        _ if interface.eq("kv")
```
Mossaka, line 33: I wonder if there is a way to avoid hard cording all the interface names here, and below. I don't know the answer myself :( Will need to think about it. Otherwise everything lgtm!

Mossaka (approved):

slight/src/commands/add.rs
danbugs: Yeah~ It's annoying 'cause every time that we add a new interface or a new dependency to an interface, we'll have to modify the `add` command...

In any case, considering this is just a "hack" of a more solidified `add` command for when the wasm registry is more fleshed out, I figured it would be alright.

PR #216, opened 2022-10-06
updating README getting started exp + other small doc changes

danbugs:

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ In simple terms, SpiderLightning allows you to go:
 
 
 ## Repository Structure
+
 - `/wit`: the SpiderLightning specification written in `*.wit` format (see [WIT](https://github.com/bytecodealliance/wit-bindgen/blob/main/WIT.md))
 - `/src`: the SpiderLightning host cli (i.e., Slight)
 - `/crates`: service implementations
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -53,8 +81,9 @@ $ make build-c # builds our c example
 $ make run-c # runs our c example
 ```
 
-> Note: There are also Rust examples that can be built (`build-rust`) and ran (`run-rust`). However, some of these examples have some dependencies on environment variables or local programs (e.g., `etcd`), so it probably won't work right off the gate like the C one did.
+# Building Rust Examples
 
+There are also Rust examples that can be built (`build-rust`) and ran (`run-rust`). However, some of these examples have some dependencies on environment variables or local programs (e.g., `etcd`), so it probably won't work right off the gate like the C one did.
```
carolynvs, line 86: Maybe just tell them that you can build them but you don't recommend running because it requires additional setup that we don't have documented yet (or plan to never document)?

Mossaka (approved):
lgtm!

carolynvs (commented):
lgtm

README.md
danbugs: Thank you for your review, Carolyn!

PR #219, opened 2022-10-14
added support for macos-latest aarch64

danbugs:


Makefile
```diff
@@ -150,12 +148,16 @@ run-restaurant-backend:
 ### GITHUB RELEASES
 .PHONY: prepare-release
 prepare-release:
-	tar -C target/ -czvf slight-unix.tar.gz release/slight
+	tar -C target/ -czvf slight-ubuntu.tar.gz release/slight
```
Mossaka, line 151: shouldn't this be `slight-linux-arm64.tar.gz`?

I would recommend put semantic version in the name.

Makefile
```diff
@@ -150,12 +148,16 @@ run-restaurant-backend:
...
 	tar -C templates/ -czvf rust-template.tar.gz rust
 	tar -C templates/ -czvf c-template.tar.gz c
 
 .PHONY: prepare-release-win
 prepare-release-win:
-	tar -C target/ -czvf slight-win.tar.gz release/slight.exe
+	tar -C target/ -czvf slight-windows.tar.gz release/slight.exe
```
Mossaka, line 157: same comment as above. `slight-windows-x86_64`

.github/workflows/release.yml
```diff
@@ -49,26 +59,61 @@ jobs:
...
+        run: tar -C target/ -czvf slight-macos-aarch64.tar.gz aarch64-apple-darwin/release/slight
+        if: ${{ fromJSON(matrix.config.os == 'macos-latest') && 
+          fromJSON(matrix.config.arch == 'aarch64') }}
+
+      - name: "add wits, and slight-ubuntu and templates tars to release"
         uses: softprops/action-gh-release@v1
         with:
           files: |
-            slight-unix.tar.gz
+            slight-ubuntu.tar.gz
```
Mossaka, line 93: same comment as below

.github/workflows/release.yml
```diff
@@ -49,26 +59,61 @@ jobs:
...
           OPENSSL_DIR: C:\Program Files\OpenSSL-Win64\
+          OPENSSL_NO_VENDOR: true
+
+      - name: "build for macos amd64"
+        run: make build
+        if: ${{ fromJSON(matrix.config.os == 'macos-latest') && 
+          fromJSON(matrix.config.arch == 'amd64') }}
+      
+      - name: "build for macos aarch64"
+        run: cargo build --manifest-path ./slight/Cargo.toml --release ${{ matrix.config.extraArgs }}
```
Mossaka, line 70: could you please put the command in makefile?

.github/workflows/ci.yaml
```diff
@@ -88,16 +94,21 @@ jobs:
...
       - name: "Build SpiderLightning and Slight"
         run: make build
-        if: ${{ fromJSON(matrix.config.os == 'macos-latest') }}
-        env:
-          OPENSSL_ROOT_DIR: /usr/local/opt/openssl
+        if: ${{ fromJSON(matrix.config.os == 'macos-latest') && 
+          fromJSON(matrix.config.arch == 'amd64') }}
+
+      - name: "Build SpiderLightning and Slight"
+        run: cargo build --manifest-path ./slight/Cargo.toml --release ${{ matrix.config.extraArgs }}
```
Mossaka, line 101: same comment as below

Mossaka (commented):
lgtm! had some nitpicking comments

Makefile
danbugs: done ✅

Makefile
danbugs: done ✅

.github/workflows/release.yml
danbugs: done ✅

.github/workflows/ci.yaml
danbugs: ~didn't want to do it 'cause of the github action specific argument

.github/workflows/release.yml
danbugs: ~didn't want to do it 'cause of the action specific argument

.github/workflows/ci.yaml
```diff
@@ -19,14 +19,18 @@ jobs:
       matrix:
         config:
           - os: "ubuntu-latest"
-            extension: ""
             platformAgnosticChecks: true
-          - os: "macos-latest"
-            extension: ""
-            extraArgs: ""
+          - {
```
Mossaka, line 23: stupid question: do we really need curly braces to define a map of key value pairs?

.github/workflows/ci.yaml
```diff
@@ -71,6 +76,7 @@ jobs:
         run: |
           rustup target add wasm32-wasi
           rustup target add wasm32-unknown-unknown
+          rustup target add aarch64-apple-darwin
```
Mossaka, line 79: nitpicking: this target is only needed in macos worker.

.github/workflows/ci.yaml
Mossaka: sounds reasonable. My only concern is that if we refactor the code in the future that requires us to change the `cargo build` command to some other path, we might forget about changing the command in the github action.

Mossaka (approved):

PR #222, opened 2022-10-18
v0.2.0 (`slightfile` redesign)

danbugs:
This includes the `slightfile` redesign.

P.S. It is not backwards compatible -- We might want to create an issue to do this before the next release to make sure it is, I didn't do it here because it felt like the PR would get too big.

### DIFF
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -17,11 +17,15 @@ use slight_http_api::HttpHandlerData;
 ///
 /// It contains:
 ///     - a `resource_map`,
+///     - a `implementor`,
+///     - a `name`,
 ///     - a `configs_map`, and
 ///     - the `slightfile_path`.
 #[derive(Clone, Default)]
 pub struct BasicState {
     pub resource_map: ResourceMap,
+    pub implementor: String,
+    pub name: String,
     pub configs_map: Option<HashMap<String, String>>,
     pub slightfile_path: PathBuf,
 }
@@ -29,14 +33,24 @@ pub struct BasicState {
 impl BasicState {
     pub fn new(
         resource_map: ResourceMap,
+        implementor: String,
+        name: String,
         configs_map: Option<HashMap<String, String>>,
         slightfile_path: impl AsRef<Path>,
     ) -> Self {
         Self {
             resource_map,
+            implementor,
+            name,
             configs_map,
             slightfile_path: slightfile_path.as_ref().to_owned(),
         }
+    }
+}
+
+impl std::fmt::Debug for BasicState {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "BasicState {{name: {}, implementor: {}}}", self.name, self.implementor)
     }
 }
 
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -1,4 +1,4 @@
-use std::path::{Path, PathBuf};
+use std::{path::{Path, PathBuf}, collections::HashMap};
 
 use anyhow::Result;
 use as_any::AsAny;
@@ -14,24 +14,24 @@ use slight_http_api::HttpHandlerData;
 ///
 /// It contains:
 ///     - a `resource_map`,
-///     - a `secret_store`, and
+///     - a `configs_map`, and
 ///     - the `slightfile_path`.
 #[derive(Clone, Default)]
 pub struct BasicState {
     pub resource_map: ResourceMap,
-    pub secret_store: String,
+    pub configs_map: Option<HashMap<String, String>>,
     pub slightfile_path: PathBuf,
 }
 
 impl BasicState {
     pub fn new(
         resource_map: ResourceMap,
-        secret_store: &str,
+        configs_map: Option<HashMap<String, String>>,
         slightfile_path: impl AsRef<Path>,
     ) -> Self {
         Self {
             resource_map,
-            secret_store: secret_store.to_string(),
+            configs_map,
             slightfile_path: slightfile_path.as_ref().to_owned(),
         }
     }
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -1,4 +1,7 @@
-use std::{path::{Path, PathBuf}, collections::HashMap};
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+};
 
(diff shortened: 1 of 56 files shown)


slight/src/commands/run.rs
```diff
@@ -111,82 +134,163 @@ async fn shutdown_signal() {
...
-                        slight_builder =
-                            slight_builder.add_state(State::Kv(slight_kv::KvState::new(
-                                resource_type.to_string(),
-                                BasicState::new(resource_map.clone(), ss, &toml_file_path),
-                            )))?;
+                        linked_capabilities.insert("kv".to_string());
+                    }
+
+                    if let std::collections::hash_map::Entry::Vacant(e) =
+                        capability_store.entry(c.name.clone())
```
Mossaka, line 169: I would extract line 168 - 180 to a function, and reuse it in other capabilities.

crates/runtime-configs/src/lib.rs
```diff
@@ -194,19 +200,43 @@ pub async fn set(
...
-                    config_name, state.secret_store
+                    "failed to get '{}' secret using secret store type: '{}'",
+                    config_name, store
                 )
             })?,
     )?;
     Ok(config)
 }
+
+fn maybe_get_config_store_and_value(c: &str) -> Result<(String, String)> {
```
Mossaka, line 230: This function is pretty hard to understand. Could you write some unittests for it?

crates/runtime-configs/src/lib.rs
```diff
@@ -139,6 +141,7 @@ impl slight_events_api::Watch for ConfigsInner {}
 /// As per its' usage in `ConfigsInner`, it must `derive` `Debug`, and `Clone`.
 #[derive(Debug, Clone)]
 pub enum ConfigsImplementor {
+    Local,
```
Mossaka, line 144: What are the usages of this Local config implementator?

crates/common/src/lib.rs
```diff
@@ -14,29 +17,47 @@ use slight_http_api::HttpHandlerData;
...
+///     - a `name`,
+///     - a `configs_map`, and
 ///     - the `slightfile_path`.
 #[derive(Clone, Default)]
 pub struct BasicState {
     pub resource_map: ResourceMap,
-    pub secret_store: String,
+    pub implementor: String,
+    pub name: String,
+    pub configs_map: Option<HashMap<String, String>>,
```
Mossaka, line 29:
```suggestion
    pub configs_map: HashMap<String, String>,
```
because HashMap has a empty default.

slight/src/commands/run.rs
```diff
@@ -111,82 +134,163 @@ async fn shutdown_signal() {
...
+    linked_capabilities: &mut HashSet<String>,
+    capability_store: &mut HashMap<String, BasicState>,
     resource_map: Arc<Mutex<StateTable>>,
     module: impl AsRef<Path>,
 ) -> Result<Builder> {
     let mut builder = Builder::new_default(module)?;
     let mut slight_builder = SlightCtxBuilder::default();
     builder.link_wasi()?;
-    if toml.specversion.as_ref().unwrap() == "0.1" {
+    if toml.specversion.as_ref().unwrap() == "0.2" {
```
Mossaka, line 145: If we do not want to support backward compatibility, I would suggest we keep specversion at 0.1 for now?

slight/src/commands/run.rs
```diff
@@ -33,32 +34,54 @@ pub async fn handle_run(module: impl AsRef<Path>, toml_file_path: impl AsRef<Pat
...
+    let http_enabled = caps.iter().any(|cap| cap.resource == "http");
 
     if events_enabled {
         log::debug!("Events capability enabled");
-        let guest_builder =
-            build_store_instance(&toml, &toml_file_path, resource_map.clone(), &module)?;
+        let guest_builder = build_store_instance(
+            &toml,
+            &toml_file_path,
+            &mut linked_capabilities,
```
Mossaka, line 66: If you pass the mutable reference of `linked_capabilities` and `capability_store` to events and http buidler, that will effectively make it link none of the capabilities, because the hashset is not empty after line 44. How about `build_store_instance` creates `linked_capability` and `capability_store` variables locally?

Mossaka (changes requested):
This is great~

slight/src/commands/run.rs
danbugs: done ✅

crates/runtime-configs/src/lib.rs
danbugs: done ✅

crates/runtime-configs/src/lib.rs
danbugs: ~local is for configs that are defined directly in the `toml` file. Here's an example: 

```toml
specversion = "0.1"
[[capability]]
resource = "kv.azblob"
name = "customers"
  [capability.configs]
  c = "world"
```

slight/src/commands/run.rs
danbugs: We do want to support backwards compatibility -- It's just not here yet. I was suggesting we could merge this as is, but, before triggering the `v0.2.0` release, make another PR TO ensure it's there. Do you think I should make it backwards compatible right off the gate?

crates/common/src/lib.rs
danbugs: discussed in chat ✅

slight/src/commands/run.rs
danbugs: done ✅

crates/runtime-configs/src/lib.rs
Mossaka: Can you mix them?
```toml
[capability.configs]
c = "world"
d = "${azapp.d}"
```

slight/src/commands/run.rs
Mossaka: No, I imagine it will take some time to support the old specversion. I suggest we don't support the old one anymore, and keep specversion at `0.1`

danbugs: Actually, I just made it backwards compatible 😆

crates/runtime-configs/src/lib.rs
danbugs: Yep~ You can mix them. They are all individually parsed.

Mossaka (approved):
lgtm!

PR #240, opened 2022-11-08
refactor: runtime takes no dependencies on capabilities

Mossaka:
I have to admit that this is a relative large PR and apologize beforehand. It refactors the core of slight - the runtime crate to not depend directly on any of the slight capabilities, including slight-kv, slight-mq, etc. This comes at a cost though. The slight CLI implementation now needs to know how to constrct a capability service before adding it to the slight runtime builder, which is a acceptable compromise. 

In addition to the runtime refactoring, this PR also cleans up some redudent code in slight capabilities. Particularly, this PR removed capability-state. 


squillace:
@Mossaka: I'd like to know what the perf difference is at startup. When you say there's a tradeoff, I wanna know _what_ it is....

crates/runtime/src/lib.rs
```diff
@@ -52,23 +60,24 @@ pub struct Builder {
     linker: Linker<Ctx>,
     engine: Engine,
     module: Module,
-    states_builder: SlightCtxBuilder<Self>,
+    // states_builder: SlightCtxBuilder<Self>,
```
devigned, line 63: Please remove.

slight/src/commands/run.rs
```diff
@@ -6,17 +6,18 @@ use std::{
...
 use slight_events_api::StateTable;
 use slight_http::Http;
 use slight_kv::Kv;
 use slight_lockd::Lockd;
 use slight_mq::Mq;
 use slight_pubsub::Pubsub;
 use slight_runtime::{
-    ctx::{SlightCtxBuilder, State},
-    Builder, Ctx,
+    // ctx::{SlightCtxBuilder, State},
```
devigned, line 18: Please remove.

devigned (changes requested):
Definitely a wide reaching PR, but one that I'm glad to see materialize. I've only added some trivial comments about removing some commented code. Outside of that, it lgtm!

devigned:
> @Mossaka: I'd like to know what the perf difference is at startup. When you say there's a tradeoff, I wanna know _what_ it is....

I believe the tradeoff is not runtime impacting, but rather code design impacting. It makes the CLI need to know a bit more about the capabilities. I believe this is a preferable tradeoff when weighed against the runtime needing to know details about the capabilities.

squillace:
OK, perfect. I saw the internal chatting after commenting. Sounds great to me!

crates/runtime/src/lib.rs
```diff
@@ -141,3 +144,29 @@ pub fn default_wasi() -> Result<WasiCtx> {
...
+
+        builder
+            .link_wasi()?
+            .link_capability::<Kv>()?
+            .add_to_builder("kv".to_string(), kv);
+
+        let (_, _) = builder.build().await;
+        Ok(())
+    }
+}
```
danbugs, line 172: I think it might make sense to move this entire test somewhere else and remove the dependency on `slight_kv` as well. What are your thoughts?

slight/src/commands/run.rs
```diff
@@ -147,132 +147,87 @@ fn build_store_instance(
...
-                    builder.link_capability::<Http<Builder>>(resource_type.to_string())?;
-                    linked_capabilities.insert("http".to_string());
-
-                    slight_builder = slight_builder.add_state(State::Http(
-                        slight_http::HttpState::<Builder>::new(resource_map.clone()),
-                    ))?;
+                    let http = slight_http::Http::<Builder>::new(resource_map.clone());
+                    builder
+                        .link_capability::<Http<Builder>>()?
+                        .add_to_builder("http".to_string(), http);
```
danbugs, line 230:
```suggestion
                    let http = slight_http::Http::<Builder>::new(resource_map.clone());
                    builder
                        .link_capability::<Http<Builder>>()?
                        .add_to_builder("http".to_string(), http);
                    linked_capabilities.insert("http".to_string());
```

slight/src/commands/run.rs
```diff
@@ -147,132 +147,87 @@ fn build_store_instance(
...
-                    &toml_file_path,
-                )?;
-
-                slight_builder = slight_builder.add_state(State::Kv(slight_kv::KvState::new(
-                    resource_type.to_string(),
-                    capability_store.clone(),
-                )))?;
+                let resource =
+                    slight_kv::Kv::new(resource_type.to_string(), capability_store.clone());
+                builder.add_to_builder("kv".to_string(), resource);
```
danbugs, line 181: The previous alternative to this (i.e., `maybe_add_named_capability_to_store`) was performing some error handling - like, bailing when two capabilities of the same name are added, and whatnot. Is any of that present now?

danbugs (changes requested):
mostly lgtm~ There's one thing I think you missed, and I want to address the possible removal of some error handling that we had previously before merging this in.

crates/runtime/src/lib.rs
Mossaka: The dependency on `slight-kv` is only in `dev-dependencies`, so when we release the crate with `--release` mode, it won't actually be compiled to the package. Thus I am okay with having a slight-kv dev-dependency for testing purposes.

slight/src/commands/run.rs
Mossaka: wooo I forgot adding back this line! Thanks

slight/src/commands/run.rs
Mossaka: I believe so. I moved that function invocation to above https://github.com/deislabs/spiderlightning/pull/240/files#diff-81fdd571c3ec358ab5f1a5984fd30948d9db1ecfa8d5ff2f016e5a412a9f7d35R146-R153

arschles (commented):
@Mossaka I'm not too familiar with this codebase but overall, looks good to me. I think some documentation on at least the newly introduced `pub` types would be good.

crates/http/src/lib.rs
```diff
@@ -226,7 +208,7 @@ impl<T: Buildable + Send + Sync + 'static> http::Http for Http<T> {
         router: &Self::Router,
     ) -> Result<Self::Server, Error> {
         // Shared states for all routes
-        let instance_builder = self.host_state.builder.as_ref().unwrap().clone();
+        let instance_builder = self.builder.as_ref().unwrap().clone();
```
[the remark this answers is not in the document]
Mossaka: I thought ? can only be implemented to `Result` type. `as_ref()` returns an Option instead

danbugs (approved):
LGTM~ cool workspace feature usage! ;D

devigned (approved):
LGTM! Great work, @Mossaka!

PR #243, opened 2022-11-11
refactor: removed events capability

Mossaka:
Removing events capability for now since we haven't figured out how to combine pubsub and events.


danbugs (commented):
I was under the impression we wanted to keep this in until using `pubsub` as event handling was more solidified. Do we want to maybe keep this as a draft until then?

Mossaka:
> I was under the impression we wanted to keep this in until using pubsub as event handling was more solidified. Do we want to maybe keep this as a draft until then?

My intention is to remove code for events for now. Later if we decide on how we deal with `pubsub` and `events`, we can always bring back the code if we want.

danbugs (approved):

Issue #249, opened 2022-11-22
create spiderlightning-pubsub interface proposal (mention slightfile config additions)


danbugs:

danbugs:
~ see: https://github.com/deislabs/spiderlightning/pull/259#issuecomment-1338031976

Issue #247, opened 2022-11-22
pubsub re-design


danbugs:
- [x] #248
- [x] ~#249~
- [x] #250
- [x] ~#251~
- [x] #252
- [x] #253
- [x] ~#254~
- [x] #272

> Note: moved #254 outside of the epic, as it will be a more involved issue on its' own.

PR #259, opened 2022-11-25
start of the pubsub proposal

danbugs:
This is part of #247 and is addressing #249

You can see the rendered version of the file: [here](https://github.com/danbugs/spiderlightning/blob/danbugs/pubsub-proposal/proposals/pubsub-mq-interfaces-proposal.md).


proposals/pubsub-mq-interfaces-proposal.md
```diff
@@ -0,0 +1,144 @@
...
+	numbering=true
+	autoSave=true
+	/vscode-markdown-toc-config -->
+<!-- /vscode-markdown-toc -->
+
+##  2. <a name='Background'></a>Background
+
+###  2.1. <a name='message-queueBackground'></a>`message-queue` Background
+
+`message-queue`s are a messaging mechanism, where message transmitters send messages to a queue construct that can be read by receivers.
```
Mossaka, line 33: should mention that messages stored in a message queue can only be consumed once by one consumer.

proposals/pubsub-mq-interfaces-proposal.md
```diff
@@ -0,0 +1,144 @@
...
+
+###  2.3. <a name='pubsubmessage-queueComparison'></a>`pubsub`/`message-queue` Comparison
+
+Historically, `pubsub` services (i.e., implementers of the `pubsub` pattern)  are thought to be comprised of non-competitive consumers – That is, if a message is sent to a channel, every receiver that is subscribed to that channel will, eventually, receive it.
+
+While, on the other hand, historically, `message-queue`s are thought to be comprised of competitive consumers – That is, a queued message is dequeued as soon as any one receiver has read it.
+
+However, if we look at services that provide `message-queue` / `pubsub` implementations, the line between the two concepts becomes blurred.
+
+That is, today's most used related services provide both `message-queue`, and `pubsub` capabilities. Below, is a visual example from [HiveMQ](https://www.hivemq.com/) displaying an event-driven architecture with both `message-queue`, and `pubsub` capabilities:
```
Mossaka, line 60: This section is really talking about the comparison between "topics" and "queues". PubSub is a pattern for asynchronous communication between publishers and subscribers. It does not assume whether the message is sent to one receiver or multiple receivers. Topics and queues, however, are terms used to describe non-competitive vs competitive-consumers. A message in a topic is sent, at the same time, to all subscribes to that topic whereas a message in a queue is sent once to one consumer.

Mossaka: Topics = fan-out
queues = load-balancing

proposals/pubsub-mq-interfaces-proposal.md
```diff
@@ -0,0 +1,144 @@
...
+
+![](./images/pubsub-interface-proposal-0.gif)
+Source: [HiveMQ's Shared Subscriptions](https://www.hivemq.com/docs/hivemq/4.9/user-guide/shared-subscriptions.html)
+
+##  3. <a name='pubsubhttpInteroptability'></a>`pubsub`/`http` Interoptability
+
+`pubsub`s are often used in conjuction with `http` to send, and receive messages from subscribed topics. However, through experimentation on the intereoperability between the two services, it is impossible to use `slight`'s `http` service to correctly manage `pubsub` subscriptions. This is because the `http` interface cannot return a handle to the `pubsub` client under which the subscription was created - Hence, you can only receive messages from a subscribed topic if the subscriptions is done in conjunction with the message polling. This is a problem because message polling, and subscribing should be two separate concerns as, if you only subscribe to a topic at the time of polling, you will miss messages that were sent to the topic before you subscribed to it.
```
Mossaka, line 67: Probably should mention that the underlying limitation is that each http handler is an independent wasm module, meaning there is no global variables to share among different http handlers. Thus, there is no global handle to message broker in the http handler.

proposals/pubsub-mq-interfaces-proposal.md
```diff
@@ -0,0 +1,144 @@
...
+	- the `payload` type is lacking, and
+	- `http` interoptability is lacking - It should be modified to allow better subscription management.
+
+##  5. <a name='Proposal'></a>Proposal
+
+The change involves:
+- unify `pub` and `sub` resources into a single `pubsub` resource,
+- modify the `pubsub` interface to contain `message-queue` semantics,
+- change the `payload` type to `event` (following the `CloudEvents` spec),
+- remove the `subscribe` method from the `pubsub` interface to pass the responsibility of subscriptions to `slight` through information made available at compilation-time through the `slightfile`, and
```
Mossaka, line 82: Probably should explain why we favor a static subscription model instead of dynamic subscription model. That is, we favor users to define what topics/queues to subscribe to in a static file, instead of in code / runtime.

proposals/pubsub-mq-interfaces-proposal.md
```diff
@@ -0,0 +1,144 @@
...
+	queue(string),
+	topic(string)
+}
+
+resource pubsub {
+	// creates a handle to a pubsub object
+	static open: func(name: string) -> expected<pubsub, error>
+
+	// unsubscribes from a channel
+	unsubscribe: func(location: channel) -> expected<unit, error>
```
Mossaka, line 101: Do we really need this method?

proposals/pubsub-mq-interfaces-proposal.md
```diff
@@ -0,0 +1,144 @@
...
+```toml
+specversion = "0.2"
+
+[[capability]]
+resource = "pubsub.mosquitto"
+name = "my-pubsub"
+    [capability.configs]
+    MOSQUITTO_HOST = "localhost"
+    MOSQUITTO_PORT = "1884"
+	SUBSCRIBE_TO = "my-subscription"
```
Mossaka, line 124: Is regex subscription pattern in-scope for this design? If not, do we plan to write another design doc for it?

Mossaka, line 124: Do we want to make `subscribe_to` and `channel_type` required arguments for pubsub configs?

Mossaka, line 124: If we combine both pub and sub, does it mean that users who want to use publisher only are forced to open a subscriber as well?

Mossaka (commented):
The first draft looks decent to me! Left some comments / questions on some design choices and pubsub concepts.

proposals/pubsub-mq-interfaces-proposal.md
danbugs: Considering subscriptions are handled by the runtime, I figured it would be a good idea to, at least, be able to unsubscribe - What are your thoughts?

proposals/pubsub-mq-interfaces-proposal.md
danbugs: I have not included it as part of this design- I don't think it needs to be another write-up when it comes, it can just be a feature addition on top of this.

danbugs: Yes~ The code should panic if they are not present.

danbugs: It depends on the crate implementation ~ for Mosquitto, there's only one client open at any time per Wasm module.

proposals/pubsub-mq-interfaces-proposal.md
danbugs: done ✅

proposals/pubsub-mq-interfaces-proposal.md
danbugs: done ✅

proposals/pubsub-mq-interfaces-proposal.md
danbugs: done ✅

danbugs:
The concluded interface (i.e., from `wasi-messaging`) very closely resembles the already existing one — w/ the only differences being:
(1) the change from `payload` to `event`, and 
(2) creation of the `channel` type to unify the `mq`, and `pubsub` capabilities.

... so I'm closing this issue, and will get to work on the implementation.

PR #260, opened 2022-12-01
chore: updated dependencies

Mossaka:
This PR updates spiderlightning dependencies to the newest. 


Makefile
```diff
@@ -36,8 +36,9 @@ install-deps:
 	sudo mv wasi-sdk-15.0/* /opt/wasi-sdk/
 	sudo rm -rf wasi-sdk-*
 	sudo apt-get update
-	sudo apt-get install lsb-release
-	sudo apt-get install -y redis-server
+	sudo apt-get install lsb-release -y
+	sudo apt-get install redis-server -y
+	sudo apt-get install protobuf-compiler -y
```
danbugs, line 41: Why are we adding this dependency?

danbugs:
I think it's best to merge this after #269 is in~ Also, I'd suggest just undoing the changes made to `azsbus`. I've applied the same changes in #270, but we're using a `TopicClient` instead.

Makefile
Mossaka: It's because the newest Redis depend on protob

danbugs (approved):
LGTM

PR #263, opened 2022-12-06
pubsub refactor: added subscription tokens to receive fxn

danbugs:
> Note: This PR includes work built on top of #261 — It should probably only be reviewed after that is merged in.

Prior to this, the `receive` function would receive one message from any topic that a client is subscribed to. This is a problem because, in addition to it, `slight` only creates one access point (client) to the messaging broker. That means that multiple users accessing `slight` through HTTP would be interfacing w/ a single client and could receive messages that are meant for one another.

The first solution I tried to solve this was to modify the `receive` function to take a topic name, and receive from a specific client. This proved to be a problem because receiving from a topic is not widely supported across implementers that we wish to target — e.g., Apache Kafka can't do it. To get close to it, you'd have to assign a specific partition (i.e., whichever the topic we are reading from is at) to a client prior to a read and manage partition offsets in subsequent reads, which is not desirable and overlaps with a possible future "partitioned messaging interface".

The second solution (i.e., the content of this PR) was to add a subscription token to the return type of a `subscribe` call — `slight` will manage that as a key in a hash map of consumers, and `receive` can use that token to get messages from a specific token via targeted consumer access.


crates/pubsub/src/implementors/mosquitto.rs
```diff
@@ -65,31 +86,93 @@ impl MosquittoImplementor {
...
+    pub async fn subscribe(&self, topic: &str) -> Result<String> {
+        let new_consumer = block_in_place(|| {
             Handle::current().block_on(async move {
-                self.client
-                    .lock()
-                    .unwrap()
-                    .subscribe(topic, QoS::AtMostOnce)
+                let mut client = Client::with_auto_id().unwrap();
+
+                client
```
devigned, line 130: I think this implies that each time subscribe is called, then a new TCP connection will be made to the broker. What do you think about creating a client per named capability, rather than per subscription? I could imagine a lot of TCP connections if it is done per subscription.

examples/pubsub-consumer-demo/src/main.rs
```diff
@@ -5,10 +5,25 @@ wit_error_rs::impl_error!(pubsub::Error);
...
+        loop {
+            let msg = ps.receive(&sub_tok)?;
+            if !msg.is_empty() {
+                println!("received message from topic 'rust'> value: {:?}", String::from_utf8(msg));
+                break;
+            }
+        }
+
+        loop {
+            let msg = ps.receive(&sub_tok1)?;
```
devigned, line 21: +1 to the token based receive.

devigned (changes requested):
I added a couple of comments outside of the review. I like the direction with this design. 

If it is not too onerous, it would be nice to see the proposal updated to reflect the changes. I think this would help other reviewers as well.

Is the message nack'd if an error is returned? I don't see the consumer explicitly ack'ing nor nack'ing.

crates/pubsub/src/implementors/mosquitto.rs
danbugs: I see two problems w/ that.

(1) Due to the fact that, in several implementors, `subscribe`s are tied to a client instance — We need a new client per subscription to be able to differentiate them and get a proper subscription token that guarantees a `receive` for that one topic. 

(2) In addition, capabilities are established at compilation time — If we follow tie subscriptions to named capabilities, we'd have to know how many subscriptions we want then, which would restrict our coverage of user applications.

danbugs:
> If it is not too onerous, it would be nice to see the proposal updated to reflect the changes. I think this would help other reviewers as well.
Yep, I'm gonna update the proposal.

> Is the message nack'd if an error is returned? I don't see the consumer explicitly ack'ing nor nack'ing.
In pull-based receives we have implicit ACKs/NACKs (i.e., as long as we don't error out, a successful return is an ACK — e.g., for Kafka that will be `commit_message`). On the other hand, when we get to push-based receives, I think we'll need an explicit ACK fxn.

crates/pubsub/src/implementors/mosquitto.rs
devigned: Ahh... I see. Thank you for pointing that out. I was looking at the mosquito implementation. After looking at the Kafka implementation your point becomes clear.

danbugs: I tried looking to see if I could change the Mosquitto implementation, but have had no luck so far. With that, is this PR good to merge for you? After this, I'm gonna move the stuff that implements `mq` to implement `pubsub` instead and complete the to `messaging` re-name  — I'm hoping that I'll get some inspiration when doing the service bus implementation 😆

devigned: Thank you for giving it a try. We'll probably have to come back to it at some point in the future, but let's not impede your work.

devigned (approved):
lgtm

PR #270, opened 2022-12-13
merging mq into messaging

danbugs:


Mossaka:
Should I wait to review this PR until #269 is merged?

Mossaka:
Could you please rebase this PR to the main branch?

crates/messaging/src/lib.rs
```diff
@@ -138,15 +149,23 @@ pub enum PubImplementor {
...
     Mosquitto(mosquitto::Pub),
+    Filesystem(filesystem::FilesystemImplementor),
+    AzSbus(azsbus::AzSbusImplementor),
 }
 
 impl PubImplementor {
-    async fn new(messaging_implementor: &str, slight_state: &BasicState) -> Self {
+    async fn new(messaging_implementor: &str, slight_state: &BasicState, name: &str) -> Self {
         match messaging_implementor {
             "messaging.confluent_apache_kafka" => {
```
Mossaka, line 159: > This comment isn ot related to this specific PR, but I think we should create an issue instead to leave it as a backlog item. 

We should use probably an enum type to do type matching instead of string matching for "message.<capability>". This will help to lower the chance of a typo in the string, and mitigate the complexity of keeping backward compatibility.

Mossaka (approved):
lgtm!

crates/messaging/src/lib.rs
danbugs: Initially, we went from str to enum and then enum to implementor, but, at some point, I think I removed the middle person~ That's a fair point, though. I'll make an issue!

PR #271, opened 2022-12-14
equalizing azsbus implemenation w/ kafka and mosquitto

danbugs:
This implementation brings the same features available in mosquitto and kafka (i.e., as part of messaging) to azsbus.

I've also updated the examples, and they all run.


crates/messaging/src/implementors/azsbus.rs
```diff
@@ -0,0 +1,102 @@
+use azure_messaging_servicebus::service_bus::SubscriptionReceiver;
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
+use tokio::runtime::Handle;
+use tokio::task::block_in_place;
+
+use anyhow::{Context, Result};
+use azure_messaging_servicebus::prelude::TopicClient;
```
Mossaka, line 8: I wonder what's the reason to use TopicClient instead of a QueueClient.

Mossaka (approved):
lgtm!

crates/messaging/src/implementors/azsbus.rs
danbugs: This is to allow subscriptions~ In the future, once we add support for a different channel type (i.e., a queue over a topic) like suggested in the `wasi-messaging` proposal, we can bring back the `QueueClient`.

PR #283, opened 2022-12-22
test: added a messaging integration test

### LINKED ISSUE #185: #37 subtask: add integration tests for messaging
see title

**Additional context**
n/a

Mossaka:
The messaging capability integration test was missing since the capability was introduced. This PR serves as a foundation of enabling integration tests for it. It adds a testing scenario where it involves three microservices:
1. a HTTP server that listens to PUT request on `/send` and publishes a message to the topic "room"
2. two consumer services that subscribes to "room" topic and sends the message to their corresponding channel.

The integration test verifies that the consumer output channels have the desired messages. 

There are some limitations to the testing though that reviewers may need to be aware of:

1. The integration test currently only test for a hard-coded mosquitto pub/sub server `/usr/local/sbin/mosquitto`. 
4. The integraiton test uses a UNIX specific crate "signal_child" that makes the test only executable on UNIX environment. 

This will close #185


danbugs:
Actually~ Before merging...

> The integraiton test uses a UNIX specific crate "signal_child" that makes the test only executable on UNIX environment.

Is there any alternative to this? I think it would be good to not leave Windows tests behind again...

Mossaka:
> Actually~ Before merging...
> 
> > The integraiton test uses a UNIX specific crate "signal_child" that makes the test only executable on UNIX environment.
> 
> Is there any alternative to this? I think it would be good to not leave Windows tests behind again...

I didn't find a great library to be able to catch the return signal on Windows.

Mossaka:
Just updated this long-overdue PR and it should be good to go :)

Issue #231, opened 2022-11-03
#231 epic: create interface for outgoing http requests


danbugs:
**Describe the solution you'd like**
This will allow service to service communication.

This is an epic. It will involve a myriad of tasks:
- [x] #232
- [x] #233
- [x] #234
- [x] #235

**Additional context**
n/a

PR #285, opened 2022-12-31
feat: added http-client capability and a reqwest implementation

### LINKED ISSUE #232: #231 subtask: alter the current http interface to include the capability to make outgoing requests OR create a new wit interface,

Mossaka:
This PR adds the much needed http-client capability that can make outbound http request from the user code. This is a relatively large PR that adds, modifies and deletes a bunch of things. To name the significant changes:

1. It adds a new http-client WIT file
2. It adds a new crate called http-client
3. It adds a new example called http-client-demo
4. It renames http capability to http-server

## The interface
The http-client interfaces looks like the following:

```go
use { request, response, http-error } from http-types

request: func(req: request) -> expected<response, http-error>
```

## The implementation

The implementation of the http-client capability uses Rust's popular http library - reqwest. 

This PR will close #232, #233, #234 and its under the epic #231.


Makefile
```diff
@@ -131,6 +132,8 @@ clean-rust:
     cargo clean --manifest-path ./examples/distributed-locking-demo/Cargo.toml & \
     cargo clean --manifest-path ./examples/messaging-producer-demo/Cargo.toml & \
     cargo clean --manifest-path ./examples/messaging-consumer-demo/Cargo.toml & \
+	cargo clean --manifest-path ./examples/http-client-demo/Cargo.toml & \
```
devigned, line 135: indentation is a bit off

examples/http-client-demo/src/main.rs
```diff
@@ -0,0 +1,38 @@
...
+    
+    let _ = Server::serve("0.0.0.0:3000", &router_with_route)?;
+    Ok(())
+}
+
+#[register_handler]
+fn handle_hello(_req: Request) -> Result<Response, HttpError> {
+    let req = crate::http_client::Request {
+        method: crate::http_client::Method::Get,
+        uri: "https://some-random-api.ml/facts/dog",
```
devigned, line 24: That is an interesting uri with equally interesting response.

devigned (changes requested):
Little formatting, but looks awesome!

examples/http-client-demo/src/main.rs
Mossaka: It is. I first found this uri in spin https://github.com/fermyon/spin/blob/main/examples/http-rust-outbound-http/src/lib.rs#L13

devigned (approved):
lgtm!

PR #305, opened 2023-01-12
sql capability

danbugs:
This PR adds a SQL capability to slight. Here's an old run of it working:
image


crates/sql/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+wit-bindgen-wasmtime = { workspace = true }
+wit-error-rs = { workspace = true }
+slight-common = { path = "../common" }
+slight-runtime-configs = { path = "../runtime-configs" }
+anyhow = { workspace = true }
+tracing = { workspace = true }
+tokio = { workspace = true }
+async-trait = { workspace = true }
+# sql.postgres deps
+postgres = { version = "0.19", optional = true }
```
Mossaka, line 19: From this description https://docs.rs/postgres/latest/postgres/index.html#implementation, it looks like to me `tokio_postgres` may be a better library to use since we are using tokio runtime?

danbugs: `postgres` is simply a wrapper, we should be ok going off of that.

Mossaka: "The client simply blocks on the futures provided by the async client." Since we are running postgres API in async function, we actually don't want the library to block on the future, no?

Mossaka (approved):
lgtm! Don't necessarily have to be in this PR, but can you add an integration test for SQL capability later?

PR #318, opened 2023-02-06
chore: small cleanup of http-server-demo

flavio:
Some small fixes to the http-server-demo:

* Avoid usage of `unwrap` when something goes wrong, instead return proper errors
* handle_bar: given no body is returned, the HTTP status code should be 204 instead of 200. This prevents `curl` from being stuck
* handle_foo: return HTTP status 200 when everything is fine, instead of 500


Mossaka (approved):
LGTM! I actually plan to make the error handling inside of HTTP handler to use `anyhow::Result` so that these function can surface the errors to the top.

PR #320, opened 2023-02-13
Fix: bump slight version to 0.3.0

Mossaka:


danbugs (approved):
LGTM. Please, delete the old tag and release and update it w/ a release after this commit.

PR #322, opened 2023-02-14
tied slight version to interface version when using slight add or slight new

danbugs:
With this change, we are adding a check to see if the installed version of slight is the same version of the interface that the user is trying to utilize when running `slight new`, or `slight add`.


Mossaka (approved):
lgtm!

Mossaka:
Could you please run `cargo clippy` and `cargo fmt` to fix the style issue?

PR #325, opened 2023-02-15
Refactor: Re-structure slight repo

Mossaka:
I've started to clean some stuff out in the slight repo. This PR is the first step which made the following changes
1. the root Cargo.toml has a crate named "SpiderLightning" but it only has a core crate that contains secret and enviornment implementations and a integration tests. I thought it would be a better idea to split it to `slight-core` crate and a slight-integration-tests` crate in the tests folder to seperate their concerns. 
2. I've moved more dependencies to the workspace
3. I've changed the path to tests wasm modules in the slight-integration-tests crate. 

This PR can wait for the release of 0.3.1


danbugs (approved):
LGTM

PR #333, opened 2023-02-22
Chore: bump slight version to 0.4.0

Mossaka:
This PR bumps the slight CLI to version 0.4.0. 

Do we have a story of when to bump SpiderLightning capability versions, like wasi-keyvalue? They are still at `v0.1.0` now.


danbugs (approved):
LGTM

PR #336, opened 2023-02-25
Bug: Fixed typos in `add` command and added integration tests

Mossaka:
This PR fixed a few typos in the `Add` command which causes failures to download `http-server` and `http-client` WIT interfaces. 

The issue was that the handle_add function uses kebab name for `http-server` and `http-client` but the `get_interface_downloads_by_name` uses snake name for them. This mismatch causes the function to be unsuable for downloading these two interfaces. 

This PR also adds some tracing statements into the Add command and it looks like this:
image

In addition, it adds an extra check statement `!resp.status().is_success()` to check whether the request interface returns a status code of success, which prevents users to download non-existing interfaces. Previously they will silently succeed and download interfaces which container "not found" as the only content. Now these requests will be rejected with an error saying the the requests interfaces could not be found.

Lastly, it adds one integration tests on `slight add` command to do sanity checks.

### DIFF
--- a/src/commands/add.rs
+++ b/src/commands/add.rs
@@ -3,7 +3,7 @@ use std::{
     io::{self, ErrorKind},
 };
 
-use anyhow::Result;
+use anyhow::{bail, Result};
 use slight_core::interface_parser::InterfaceAtRelease;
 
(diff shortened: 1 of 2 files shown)


danbugs (approved):
LGTM

devigned (approved):
lgtm!

PR #341, opened 2023-02-28
Fix team name typo

fcabrera23:
The team name is incorrect, so the GitHub API request will fail to get the latest release tag, and installation will fail. Changing to correct team name and verified installation using Win10/11


Mossaka (approved):
lgtm!

PR #342, opened 2023-03-02
enable handle_run to accept io redirect paths

devigned:
Currently, `handle_run` does not enable the caller to specify I/O redirection for modules running in `slight`. This PR allows the caller of `handle_run` to specify paths for std{out, in, err} which will be used to redirect WASI stdio.

- updated the clap dependency due to https://github.com/clap-rs/clap/issues/4733 and fixed related lint issues

Out of scope for the PR:
- Adding command line args to `slight`


devigned:
Going to give this a bit more testing. I'm running into 502 errors when running this in the containerd-shim.

crates/runtime/src/lib.rs
```diff
@@ -55,6 +67,7 @@ pub struct Builder {
     engine: Engine,
     module: Module,
     state_builder: SlightCtxBuilder,
+    io_redirects: Option<IORedirects>,
```
Mossaka, line 70: Would it be better to add `WasiCtx` as a field to the Builder so that users can pass any configuration of `WasiCtx` struct to slight runtime?

src/commands/run.rs
```diff
@@ -58,15 +59,27 @@ const CONFIGS_HOST_IMPLEMENTORS: [&str; 3] =
 #[cfg(feature = "sql")]
 const SQL_HOST_IMPLEMENTORS: [&str; 1] = ["sql.postgres"];
 
-pub async fn handle_run(module: impl AsRef<Path>, toml_file_path: impl AsRef<Path>) -> Result<()> {
+pub type IORedirects = slight_runtime::IORedirects;
+
+#[derive(Clone, Default)]
+pub struct RunArgs {
+    pub module: PathBuf,
+    pub slightfile: PathBuf,
```
Mossaka, line 67: I think slightfile has already been included as a global argument to `slight` CLI.

danbugs: We have the `slightfile` being pased via the `toml_file_path` arg currently~ I think David is just grouping it all together w/ `RunArgs`.

devigned: Yes, that is correct. Rather than continuing to expand the function signature, we can add fields to the `RunArgs` struct as needed.

Mossaka: Okay, slightfile arg is being used in `slight secret` command as well. Just want to make sure that it will continue to work.

Mossaka:
Let me know if this is ready to review again @devigned

crates/runtime/src/lib.rs
devigned: Perhaps, I'm just not clear on how that would look. However, it would be something I'd like to explore in a future PR.

devigned:
> Let me know if this is ready to review again @devigned

I think this last push should satisfy the linter. Should be good to go.

src/commands/test/slightfile.toml
```diff
@@ -0,0 +1,5 @@
+specversion = "0.2"
+
+[[capability]]
```
devigned, line 3: @danbugs, I thought it was interesting that if I didn't include a capability, slight would fail when I ran `handle_run`. Is that a bug? Must one use at least 1 capability?

devigned:
@danbugs and @Mossaka, can I get another review?

crates/runtime/src/lib.rs
Mossaka: Could you please raise an issue to create a backlog item for this? After that, feel free to resolve this convo.

Mossaka, line 70: I don't know if this could simplify things, but if you define `IORedirects` as a struct of `Option<PathBuf>` for each std-io, this will give granularity to users to specify only one of the IO path. Then you can say `io_redirects` is `IORedirects` instead of `Option<IORedirects>` and it could be defaulted to a struct with Nones.

crates/runtime/src/lib.rs
```diff
@@ -70,9 +83,16 @@ impl Builder {
             engine,
             module,
             state_builder: SlightCtxBuilder::default(),
+            io_redirects: None,
         })
     }
 
+    /// Set the I/O redirects for the module
+    pub fn set_io(mut self, io_redirects: IORedirects) -> Self {
```
Mossaka, line 91: Not related to this PR but something we might want to explore futher in the future is that if we think `slight CLI` as a wrapper on `wasmtime CLI`, then obviously we want to expose all the wasmtime configurations to slight CLI as well. This may require us to build something similar to `WasiCtxBuilder` and may call it `SlightCtxBuilder` which can be exposed to the users.

crates/runtime/src/lib.rs
```diff
@@ -97,14 +117,27 @@ impl Builder {
     }
 }
 
+fn maybe_open_stdio(pipe_path: &Path) -> Option<File> {
```
Mossaka, line 120: Can you write at least one unit test for this func?

crates/runtime/src/lib.rs
```diff
@@ -97,14 +117,27 @@ impl Builder {
...
+    if pipe_path.as_os_str().is_empty() {
+        None
+    } else {
+        Some(
+            OpenOptions::new()
+                .read(true)
+                .write(true)
+                .open(pipe_path)
+                .expect("could not open pipe"),
```
Mossaka, line 129:
```suggestion
                .expect(format!("could not open pipe: {path}", path = pipe_path.to_str())),
```

crates/runtime/src/lib.rs
```diff
@@ -121,6 +154,31 @@ impl WasmtimeBuildable for Builder {
     }
 }
 
+fn build_wasi_context(io_redirects: Option<IORedirects>) -> Result<WasiCtx> {
```
Mossaka, line 157: Although this func is named as "build_wasi_context", it is heavily funcitoned as a unit of building io pipes as its argument suggested so. Not a big deal but I would suggest to change it to something that takes a `WasiCtxBuilder` as an argument, and add `stdout/in/err` to the builder, and name the function to `add_io_redirecfts_to_wasi_context`.

src/commands/run.rs
```diff
@@ -340,3 +353,53 @@ fn maybe_add_named_capability_to_store(
...
+    use rand::Rng;
+    use slight_runtime::IORedirects;
+    use std::fs::File;
+    use std::path::{Path, PathBuf};
+    use tempfile::tempdir;
+    use tokio::fs;
+
+    #[tokio::test]
+    async fn test_handle_run_with_io() -> anyhow::Result<()> {
+        let module = "./src/commands/test/io-test.wasm";
```
Mossaka, line 369: I would recommend putting test artifacts to some places like `CARGO_TARGET_DIR` https://doc.rust-lang.org/cargo/reference/environment-variables.html

Mossaka: No need to change it to this PR though

src/commands/run.rs
```diff
@@ -340,3 +353,53 @@ fn maybe_add_named_capability_to_store(
...
+        let canary: String = rand::thread_rng()
+            .gen_ascii_chars()
+            .take(7)
+            .map(char::from)
+            .collect();
+        fs::write(&stdin_path, &canary).await?;
+        let _ = File::create(&stdout_path)?;
+        let _ = File::create(&stderr_path)?;
+
+        let args = RunArgs {
```
Mossaka, line 388: It would be great to see an integration test using `slight run` command and passes in io redirects paths

src/commands/test/slightfile.toml
Mossaka: File an issue for this please!

tests/io-test/slightfile.toml
```diff
@@ -0,0 +1 @@
+specversion = "0.2"
```
Mossaka, line 1: unused file?

Mossaka (commented):
Nice work! Left some comments

crates/runtime/src/lib.rs
Mossaka: `fn add_io_redirecfts_to_wasi_context(mut WasiCtxBuilder) -> Result<WasiCtxBuilder>`
`fn build_wasi_context(mut WasiCtxBuilder) -> Result<WasiCtx>`

split the logic to these two functions and use the [Builder pattern](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html)

crates/runtime/src/lib.rs
devigned: Sounds great. I think you mentioned something like this before. We should chat about what it would look like / behave.

src/commands/run.rs
devigned: In the PR description, I left this as out of scope. I wasn't sure if it would be useful for anyone or if it would just pollute the command line args.

src/commands/test/slightfile.toml
devigned: Added.

tests/io-test/slightfile.toml
devigned: Not really. I've kept it around to provide a working set of files local to the io-test slight app. I'd like to keep it around until #355 is resolved.

src/commands/run.rs
devigned: Let's discuss how we'd like to build unit test wasm apps. I think we need to include them in the build.rs or something to ensure they are up-to-date when running the unit tests.

/cc @danbugs

crates/runtime/src/lib.rs
devigned: I think I'd like to keep the optional on IORedirects as it really is optional for args. However, I did make the paths within the struct as optional as well. This provides the ability to specify some, but not all.

crates/runtime/src/lib.rs
devigned: Done, but I needed to comment out the other test in the runtime. However, when running make test, it didn't seem like it was running anyway.

I've marked a TODO to re-enable the test.

devigned:
@Mossaka I've updated the PR based on your feedback. Ready for the next round of reviews.

crates/runtime/src/lib.rs
Mossaka: Created an issue #358 for that. Will resolve this convo

crates/runtime/src/lib.rs
Mossaka: Yes, remove the `test = false` from Cargo.toml

Mossaka (approved):
LGTM! Again, thank you @devigned

PR #343, opened 2023-03-03
`slight` now pre-opens the `.` dir to allow filesystem access

danbugs:
Prior to the change, we'd get:
image

Now, w/ ambient authority, the guest app is able to read and write to the filesystem.


crates/runtime/src/lib.rs
```diff
@@ -132,7 +132,10 @@ pub fn default_config() -> Result<Config> {
 
 // TODO (Joe): expose the wasmtime wasi context as a capability?
 pub fn default_wasi() -> Result<WasiCtx> {
-    let ctx: WasiCtxBuilder = WasiCtxBuilder::new().inherit_stdio().inherit_args()?;
+    let ctx: WasiCtxBuilder = WasiCtxBuilder::new()
```
Mossaka, line 135: I made the comment in #342 that we might want to consider allowing the CLI to pass a `WasCtx` to the slight runtime.

Mossaka (approved):

crates/runtime/src/lib.rs
danbugs: ~as new command-line args or part of the slightfile? I'll merge this for now to accommodate AKS EE, but we can change it later.

PR #359, opened 2023-03-14
feat(http-server): Add WASI reactor support to http-server cap

Mossaka:
This PR switches the WASI command model to [reactor model](https://github.com/WebAssembly/WASI/issues/13#:~:text=A%20Command%20has%20a%20%22main,as%20callbacks%20to%20be%20called.) for wasi-http. The reason for that is in the command model, each exported HTTP handler function would not be able to initialize properly with the WASI stdout because the host will not invoke `_start`. The reactor solves this issue.

## Breaking changes

1. rust slight applications that use http-server must use `cargo new --lib` to create the rust program, instead of `--bin`. This will create a rust library. In Cargo.toml, the following attribute is needed:

```toml
[lib]
crate-type = ["cdylib"]
```

2. Use slight provided `slight_http_server_macro::on_server_init` macro on top of the router and server setup function.

A rust slight program that uses http-server capability would end up looking like

```rust
use http_server::*;
use slight_http_handler_macro::register_handler;
use slight_http_server_macro::on_server_init;

wit_bindgen_rust::import!("../../wit/http-server.wit");
wit_bindgen_rust::export!("../../wit/http-server-export.wit");
wit_bindgen_rust::import!("../../wit/http-client.wit");
wit_error_rs::impl_error!(http_server::HttpRouterError);
wit_error_rs::impl_error!(http_client::HttpError);

#[on_server_init]
fn main() -> Result<()> {
    let router = Router::new()?;
    let router_with_route = router
        .get("/hello", "handle_hello")?
        .get("/person/:name", "handle_person_with_name")?
        .get("/foo", "handle_foo")?
        .put("/bar", "handle_bar")?
        .post("/upload", "upload")?
        .delete("/delete-file", "delete_file_handler")?
        .get("/request", "handle_request")?;

    println!("guest starting server");
    let _ = Server::serve("0.0.0.0:3000", &router_with_route)?;
    // server.stop().unwrap();
    println!("guest moving on");

    Ok(())
}
```

## A Note to Reviewers

I've tested that the http handler would work with keyvalue capability correctly. I still need to add more docs to this newly added macro and changes to the slight runtime library, but other than that, this PR is ready to review.

### DIFF
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -5,7 +5,6 @@ use std::{
 
 use anyhow::{bail, Result};
 use as_any::Downcast;
-use wit_bindgen_wasmtime::wasmtime::Store;
 use slight_common::{BasicState, Capability, Ctx as _, WasmtimeBuildable};
 use slight_core::slightfile::{Capability as TomlCapability, TomlFile};
 #[cfg(feature = "distributed-locking")]
@@ -23,6 +22,7 @@ use slight_runtime::{Builder, Ctx};
 use slight_runtime_configs::Configs;
 #[cfg(feature = "sql")]
 use slight_sql::Sql;
+use wit_bindgen_wasmtime::wasmtime::Store;
 
 #[cfg(feature = "keyvalue")]
 const KEYVALUE_HOST_IMPLEMENTORS: [&str; 8] = [
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -77,7 +77,7 @@ pub async fn handle_run(args: RunArgs) -> Result<()> {
     tracing::info!("Starting slight");
 
     let mut host_builder = build_store_instance(&toml, &args.slightfile, &args.module).await?;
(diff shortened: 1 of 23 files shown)


src/commands/run.rs
```diff
@@ -78,45 +77,53 @@ pub async fn handle_run(args: RunArgs) -> Result<()> {
...
+            .any(|cap| cap.resource.as_ref().expect("missing resource field") == "http");
+    } else {
+        bail!("unsupported toml spec version");
+    }
     // looking for the http capability.
-    let http_enabled = if cfg!(feature = "http-server") {
-        let http_enabled;
+    if cfg!(feature = "http-server") && http_enabled {
+        log::debug!("Http capability enabled");
+        update_http_states(
```
devigned, line 101: @Mossaka I needed to add the io_redirects to the HTTP instance builder as well as the host_builder.

Mossaka: sounds good!

PR #360, opened 2023-03-14
tests: Re-enable AWS integration tests

Mossaka:
This PR re-enable AWS dynamodb integration tests which were disabled in https://github.com/deislabs/spiderlightning/pull/287.


crates/keyvalue/src/implementors/awsdynamodb.rs
```diff
@@ -44,8 +45,9 @@ impl AwsDynamoDbImplementor {
     /// }
     /// ```
     pub async fn new(_slight_state: &BasicState, name: &str) -> Self {
-        let shared_config = aws_config::load_from_env().await;
-        let client = Client::new(&shared_config);
+        let region = RegionProviderChain::default_provider().or_else("us-west-2");
+        let config = from_env().region(region).load().await;
```
danbugs, line 49: I wish we could convert this to use our runtime configs like every other crate...

tests/keyvalue-test/keyvalue_awsdynamodb_slightfile.toml
```diff
@@ -11,23 +15,25 @@ name = "rand"
...
 
 [[capability]]
 resource = "keyvalue.awsdynamodb"
 name = "random2"
-    # This capability loads its' config from environment variables. It requires:
-    #   - `AWS_ACCESS_KEY_ID`,
-    #   - `AWS_SECRET_ACCESS_KEY`, and
-    #   - `AWS_REGION`.
-
+    [capability.configs]
```
danbugs, line 27: These are not really being used by the app...

danbugs (approved):

crates/keyvalue/src/implementors/awsdynamodb.rs
Mossaka: this is a great point. We may have to set the environment variables for the running process from the slightfile and use `from_env()`

Mossaka: Fixed. Could you PTAL?

PR #361, opened 2023-03-14
feat(blob-store): Add wasi-blob-store capability

Mossaka:
This PR is largely work-in-process. It's goal is to add [wasi-blob-store](https://github.com/WebAssembly/wasi-blob-store) proposal as a capability to slight. 

I will make increment addition to this PR as seperate commits for easier review. 

- [x] structure the blob-store WIT
- [x] implement AWS S3 on blob-store interface
- [x] write an example of using S3
- [x] write an integration test to fully test the APIs of blob store
- [ ] do we want to implement `available` and `close`?


danbugs (commented):
Is this PR only for adding the interface or do you plan to add an implementor (i.e., possibly move azblob)?

Mossaka:
> Is this PR only for adding the interface or do you plan to add an implementor (i.e., possibly move azblob)?

I intended to work more on this PR to add at least one implementor. Possibly I will add aws s3

Mossaka:
This PR is mostly ready for review, although there are some remaining tasks to finish before this is mergable. I will document the tasks on the PR description section.

crates/blob-store/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    async fn container_name(&mut self, self_: &Self::Container) -> Result<String, Error> {
+        Ok(self_.implementor.name().await?)
+    }
+    async fn container_info(
+        &mut self,
+        self_: &Self::Container,
+    ) -> Result<ContainerMetadata, Error> {
+        Ok(self_.implementor.info().await?)
```
arschles, line 124: `self_` makes this easy to mistake for `self`. maybe rename to `container` or `ctr`?

```suggestion
        ctr: &Self::Container,
    ) -> Result<ContainerMetadata, Error> {
        Ok(ctr.implementor.info().await?)
```

same goes for the below functions too

tests/blob-store-test/src/main.rs
```diff
@@ -0,0 +1,65 @@
...
+    assert_eq!(bucket.name()?, "slight-test-bucket");
+
+    for name in bucket.list_objects()? {
+        println!("Found object: {}", name);
+        bucket.delete_object(&name)?;
+    }
+
+    // upload 3 files
+    for i in 0..3 {
+        // let body = std::fs::read(format!("testfile{i}.txt")).expect("should have been able to read the file");
```
arschles, line 26: delete this?

```suggestion
```

tests/blob-store-test/src/main.rs
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    // read 3 files
+    for name in bucket.list_objects()? {
+        println!("Found object: {}", name);
+        let read_stream = bucket.read_object(&name)?;
+        let contents = read_stream.read(1024 * 4)?.unwrap();
+
+        // compare the contents with content on filesystem
+        // let body = std::fs::read(format!("{}", name)).expect("should have been able to read the file");
```
arschles, line 39: delete this?

```suggestion
```

tests/blob-store-test/src/main.rs
```diff
@@ -0,0 +1,65 @@
...
+        assert_eq!(body, contents);
+    }
+
+    // read one file
+    if bucket.has_object("testfile1.txt")? {
+        let read_stream = bucket.read_object("testfile1.txt")?;
+        let contents = read_stream.read(1024 * 4)?.unwrap();
+
+        // compare the content with content on filesystem
+        // let body = std::fs::read("testfile1.txt").expect("should have been able to read the file");
```
arschles, line 49: delete this?

```suggestion
```

arschles (commented):
Just some nit-picks, but overall looks great

crates/blob-store/src/implementors/aws_s3.rs
```diff
@@ -0,0 +1,254 @@
...
+/// A write stream contains a S3 client, the bucket name and a key
+#[derive(Debug, Clone)]
+pub struct S3WriteStream {
+    client: Arc<Client>,
+    bucket: String,
+    key: String,
+}
+
+impl S3Container {
+    pub async fn new(_slight_state: &BasicState, name: &str) -> Result<Self> {
```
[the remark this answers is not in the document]
Mossaka: There is no reason to keep this parameter as of right now, but we eventually need this parameter as it contains information read from the slightfile, things like env variables from the slightfile to tell how aws services can be configured. The current implemeantion relies on the host to provide the values, which is not ideal. 

I will work on fixing this.

crates/blob-store/src/implementors/aws_s3.rs
```diff
@@ -0,0 +1,254 @@
...
+            .bucket(container.clone())
+            .key(name)
+            .object_attributes(aws_sdk_s3::model::ObjectAttributes::ObjectSize)
+            .send()
+            .await?;
+        let res = ObjectMetadata {
+            name: name.to_owned(),
+            container,
+            created_at: metadata.last_modified().unwrap().as_secs_f64() as u64, // TODO: fix me
+            size: metadata.object_size() as u64,
```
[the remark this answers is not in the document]
Mossaka: Thanks!

crates/blob-store/src/lib.rs
Mossaka: Thanks!

tests/blob-store-test/src/main.rs
Mossaka: Instead of deleting them, I will write a TODO comment saying that in the future, we want to enable this to allow users to upload files from local filesystem to blob stores. There are some security constraints with wasm modules, such as that they can't access random places in the filesystem, and thus we need to think throughly of how to allow them to upload files.

wit/blob-types.wit
```diff
@@ -0,0 +1,35 @@
...
+	created-at: timestamp,
+}
+
+// information about an object
+record object-metadata {
+	// the object's name
+	name: object-name,
+	// the object's parent container
+	container: container-name,
+	// date and time the object was created
```
devigned, line 25: u64 is a date time? Is this a resource pointer or is the date time? I see the usage with a TODO comment. Can you expand on this?

wit/blob-types.wit
```diff
@@ -0,0 +1,35 @@
+// name of a container, a collection of objects.
+// The container name may be any valid UTF-8 string.
+type container-name = string
+
+// name of an object within a container
+// The object name may be any valid UTF-8 string.
+type object-name = string
+
+type timestamp = u64
```
devigned, line 9: Why wasn't time stamp used for created_at in container-metadata and object-metadata?

wit/blob-store.wit
```diff
@@ -0,0 +1,103 @@
...
+	read-object: func(name: object-name) -> expected<read-stream, error>
+  
+	// creates or replaces an object.
+	write-object: func(name: object-name) -> expected<write-stream, error>
+	
+	// retrieves an object or portion of an object, as a resource.
+	// Start and end offsets are inclusive.
+	// Once a data-blob resource has been created, the underlying bytes are held by the blobstore service for the lifetime
+	// of the data-blob resource, even if the object they came from is later deleted.
+	// get-data: func(name: object-name, start: u64, end: u64) -> expected<data-blob, error>
```
devigned, line 25: Was this purposefully commented out?

wit/blob-store.wit
```diff
@@ -0,0 +1,103 @@
...
+  
+	// creates or replaces an object with the data blob.
+	// write-data: func(name: object-name, data: data-blob) -> expected<unit, error>
```
devigned, line 28: Was this purposefully commented out?

tests/blob-store-test/testfile.txt
devigned, line 1: nit: seems odd to have one file without an index suffix.

crates/blob-store/src/read_stream.rs
```diff
@@ -0,0 +1,29 @@
...
+
+use crate::container::DynR;
+
+#[async_trait]
+pub trait ReadStreamImplementor {
+    async fn read(&self, size: u64) -> Result<Option<Vec<u8>>>;
+    async fn available(&self) -> Result<u64>;
+}
+
+impl std::fmt::Debug for dyn ReadStreamImplementor + Send + Sync {
```
devigned, line 13: nit: consider adding the `std::fmt` use statement.

crates/blob-store/src/implementors/aws_s3.rs
```diff
@@ -0,0 +1,254 @@
...
+impl ReadStreamImplementor for S3ReadStream {
+    async fn read(&self, size: u64) -> Result<Option<Vec<u8>>> {
+        // In wasi-blob-store, `read` takes a mutable buffer as an argument.
+        // I changed it to return a vector of bytes instead because as of right now, 
+        // wit-bindgen does not support generating mutable buffers. 
+        // 
+        // This is something we might want to go back and change in the future
+        // when we transform wit-bindgen v0.2.0 to the newest component model syntax.
+        // 
+        // TODO: change `read` to take a mutable buffer as an argument
```
devigned, line 204: Thank you for calling this out.

crates/blob-store/src/implementors/aws_s3.rs
```diff
@@ -0,0 +1,254 @@
...
+        Ok(key_exists)
+    }
+    async fn object_info(&self, name: ObjectNameParam<'_>) -> Result<ObjectMetadata> {
+        let container = self.name().await?;
+        let metadata = self
+            .client
+            .get_object_attributes()
+            .bucket(container.clone())
+            .key(name)
+            .object_attributes(aws_sdk_s3::model::ObjectAttributes::ObjectSize)
```
devigned, line 149: nit: consider adding the use statement for brevity.

devigned (commented):
Mostly nits and questions. Looking pretty solid.

wit/blob-types.wit
Mossaka: That was a miss from the spec. Will raise an issue and a PR to fix this.

Mossaka: Actually, the spec has been updated to use `timestamp` everywhere. I copied the older wits, will fix them

wit/blob-store.wit
Mossaka: Yes, it was purposefully commented out and I will write a comment here expanding the reason. This PR is not going to implement the `data-blob` resource as I am not sure how it will add value in addition to the `write-stream` resource which is implemented.

wit/blob-types.wit
Mossaka: It is date & time, and I implemented it by converting the date & time struct from service response to secs.

devigned: Ok, but secs from what base? Since epoch, big bang, or some other notable event?

devigned: Just to be clear, the comment above was meant half as comedy and the other half as a serious question.

Mossaka: Oh, since UNIX Epoch

devigned: Please consider calling that out in the description of that field / parameter.

Mossaka: I guess you are suggesting for a change in the wasi-blob-store spec where we can define a `date_and_time` resource which provides functions to explictly convert date and time to some other representations.

Like secs from unix epoch as a `i64`, or nano seconds as a `i32`

devigned: I didn't see the `date_and_time` resource, but any place I see a field for datetime as an numeric it (imho) deserves to have some explanation of how that is converted from a calendar datetime to an integer to be clear about the expectations of the API. Perhaps, pointing to functionality to transform from / to datetime from / to numeric.

Mossaka: Okay raised this issue to track https://github.com/WebAssembly/wasi-blob-store/issues/7

crates/blob-store/src/implementors/aws_s3.rs
```diff
@@ -0,0 +1,254 @@
...
+            .send()
+            .await?;
+        let res = ObjectMetadata {
+            name: name.to_owned(),
+            container,
+            created_at: metadata.last_modified().unwrap().as_secs_f64() as u64, // TODO: fix me
```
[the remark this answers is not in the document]
devigned: It looks like it is based on the output from the cli: https://docs.aws.amazon.com/cli/latest/reference/s3api/get-object-attributes.html

It wasn't clear to me from the documentation, but it seems reasonable that it would be.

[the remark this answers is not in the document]
Mossaka: Yes I believe so. 

What I am not sure in this impelemntation is whether or not `f64` can be just converted to `u64`. Seems like a lot of information got lost in the transition.

crates/blob-store/src/implementors/aws_s3.rs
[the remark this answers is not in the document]
Mossaka: fixed

wit/blob-store.wit
Mossaka: the same reason as above. I am not sure whether there is `write-data` and `data-blob`. Need to sort this thing out in wasi-blob-store

Mossaka:
I've merged the azure-blob implementation to this PR and unforunately this one became much harder to review... I would really appreicate if you could take a second look @devigned, @danbugs

tests/blob-store-test/src/main.rs
Mossaka: Resovled

crates/blob-store/src/write_stream.rs
```diff
@@ -0,0 +1,29 @@
+use anyhow::Result;
+
+use async_trait::async_trait;
+
+use crate::container::DynW;
+
+#[async_trait]
+pub trait WriteStreamImplementor {
+    async fn write(&self, data: &[u8]) -> Result<()>;
+    async fn close(&self) -> Result<()>;
```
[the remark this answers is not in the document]
Mossaka: `close` is not used for the same reason as above - we don't have a real `write-stream` but merely a write function that write the entire data to a blob.

crates/blob-store/src/implementors/azblob.rs
```diff
@@ -0,0 +1,216 @@
...
+    }
+    async fn delete_objects(&self, _names: Vec<ObjectNameParam<'_>>) -> Result<()> {
+        // TODO: there isn't an API in azure blob storage to do this directly
+        // if we are going to delete a lot of objects, we should use a batch delete
+        // otherwise, we run into issues with deleting half the objects and then
+        // failing
+        //
+        //
+        // followed up on https://github.com/Azure/azure-sdk-for-rust/issues/1249
+        todo!()
```
Mossaka, line 109: @devigned I am thinking that the `delete-objects` API from wasi-blob-store is a bit too specific and demanding as that it is not something every blog store would provide. For example, there isn't delete-objects API in azure blob storage rust SDK, albert azure provides batch operations for deletions. 

I am in favor of removing this API form the spec. WDYT?

Mossaka:
Could you please give another pass? @devigned , @danbugs

crates/blob-store/src/implementors/azblob.rs
Mossaka: Will keep it there

PR #374, opened 2023-03-30
dapr key value implementation

devigned:
This PR adds a [Dapr](https://dapr.io) Key / Value implementation. It stores keys and values in the Dapr state store.

containerd/runwasi and the deislabs/containerd-wasm-shims do not currently support running wasm containers in the same pod as Linux containers, the intent is to run Dapr as a DaemonSet as demonstrated in https://github.com/salaboy/dapr-ambient.

- [x] initial implementation using the Dapr state store
- [ ] unit tests for the implementation
- [ ] integration tests to exercise the implementation


Mossaka:
Could you please add the instructions to gh actions to install & setup Dapr for integration tests? Thanks!

danbugs:
~tracked by #382

PR #375, opened 2023-03-31
refactor(slightfile): move slightfile logic to its own crate

Mossaka:
this PR does some preparation work for implementing the #351 feature in slightfile. It 
1. moves toml manifest structs to its own crate
2. defines specversion and capability as enums
3. restricts resource to only supported value
4. add a bunch of unit tests

CC @danbugs for review!


crates/slightfile/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+
+impl TryFrom<String> for SecretStoreResource {
+    type Error = anyhow::Error;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        match value.as_str() {
+            "configs.azapp" => Ok(SecretStoreResource::Azapp),
+            "configs.envvars" => Ok(SecretStoreResource::Envvars),
+            "configs.usersecrets" => Ok(SecretStoreResource::Usersecrets),
+            "configs.local" => Ok(SecretStoreResource::Local),
```
danbugs, line 103: Is there a better way to do this? I think making people that add new capabilities have to come modify this is a bit of an annoying indirection.

crates/slightfile/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+    }
+}
+
+impl From<SecretStoreResource> for String {
+    fn from(value: SecretStoreResource) -> Self {
+        match value {
+            SecretStoreResource::Azapp => String::from("configs.azapp"),
+            SecretStoreResource::Envvars => String::from("configs.envvars"),
+            SecretStoreResource::Usersecrets => String::from("configs.usersecrets"),
+            SecretStoreResource::Local => String::from("configs.local"),
```
danbugs, line 115: ~same as above.

crates/slightfile/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+    #[serde(rename = "configs.envvars")]
+    ConfigsEnvvars,
+    #[serde(rename = "configs.usersecrets")]
+    ConfigsUsersecrets,
+    #[serde(rename = "distributed_locking.etcd")]
+    DistributedLockingEtcd,
+    #[serde(rename = "lockd.etcd")]
+    V1DistributedLockingEtcd,
+    #[serde(rename = "sql.postgres")]
+    SqlPostgres,
```
danbugs, line 173: ~same as above (x2).

crates/slightfile/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+            Resource::ConfigsUsersecrets => write!(f, "configs.usersecrets"),
+            Resource::DistributedLockingEtcd => write!(f, "distributed_locking.etcd"),
+            Resource::SqlPostgres => write!(f, "sql.postgres"),
+            Resource::V1KeyvalueAwsDynamoDb => write!(f, "kv.awsdynamodb"),
+            Resource::V1KeyvalueAzblob => write!(f, "kv.azblob"),
+            Resource::V1KeyvalueFilesystem => write!(f, "kv.filesystem"),
+            Resource::V1KeyvalueRedis => write!(f, "kv.redis"),
+            Resource::V1MessagingAzsbus => write!(f, "mq.azsbus"),
+            Resource::V1MessagingFilesystem => write!(f, "mq.filesystem"),
+            Resource::V1DistributedLockingEtcd => write!(f, "lockd.etcd"),
```
danbugs, line 206: ~same as above (x3).

danbugs (changes requested):
~great job w/ tests, and error messages!

I just had a few comments I think we should address before merging.

crates/slightfile/src/lib.rs
Mossaka: I don't know there is a better way to do this... I agree that people who add new capabilities will need to come here to modify slightfile but they need to do that anyway in slight CLI and capability crate.

PR #376, opened 2023-04-04
feat(slightfile): implement wildcard namespace for resources

Mossaka:
This PR implements the wildcard namespace for resources in slightfile crate. It does not, however, implement the semantics of the star syntax to any of the capabilities. This will be a follow-up.

- [x] add `*` support to slightfile
- [x] implement validation logic to slightfile
- [x] implement a selector that supports `*`
- [x] test it out for all the capabilities.


Mossaka:
Just implement an integration test for this and all the tasks are finished. I believe this PR is ready for reivew. @danbugs , @devigned

devigned:
This is a hefty PR. I've looked over it once, but probably going to need a minute to digest. 

The wildcard messaging test flows really well.

Mossaka:
> This is a hefty PR. I've looked over it once, but probably going to need a minute to digest.

@devigned I'd suggest that you pull down the changes and play around with the examples with wildcard in the slightfile. If there are any suprises, I can help to reduce them.

crates/slightfile/src/capability_store.rs
```diff
@@ -0,0 +1,177 @@
...
+/// let mut state = State;
+/// store.insert(ResourceName::Specific("keyvalue.redis"), "redis", state);
+/// store.insert(ResourceName::Any, "messaging", state);
+///
+/// assert_eq!(store.get("keyvalue.redis"), state);
+/// ```
+///
+#[derive(Debug, Clone)]
+pub struct CapabilityStore<T> {
+    inner: HashMap<CapabilityName, HashMap<ResourceName, T>>,
```
danbugs, line 34: I think `ResourceName` and `CapabilityName` can be easily confused. Because the fields in the slight file are `resource`, and `name` respectively. Why don't we just call them `Resource`/`Name`, or, `CapabilityResource`/`CapabilityName`?

crates/slightfile/src/capability_store.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+}
+
+impl<T> CapabilityStore<T> {
+    pub fn new() -> Self {
+        Self {
+            inner: HashMap::new(),
+        }
+    }
+    pub fn get(&self, name: &str, cap: &str) -> Option<&T> {
```
danbugs, line 49: ~similar comment to before, let's definitely not call `resource` `name`, when we have another field already called `name`.

crates/slightfile/src/capability_store.rs
```diff
@@ -0,0 +1,177 @@
...
+/// use slight_file::slightfile::ResourceName;
+///
+/// pub struct State;
+///
+/// let mut store = CapabilityStore::new();
+/// let mut state = State;
+/// store.insert(ResourceName::Specific("keyvalue.redis"), "redis", state);
+/// store.insert(ResourceName::Any, "messaging", state);
+///
+/// assert_eq!(store.get("keyvalue.redis"), state);
```
danbugs, line 29: Doesn't get take two strings?

crates/slightfile/src/capability_store.rs
```diff
@@ -0,0 +1,177 @@
...
+}
+
+impl<T> CapabilityStore<T> {
+    pub fn new() -> Self {
+        Self {
+            inner: HashMap::new(),
+        }
+    }
+    pub fn get(&self, name: &str, cap: &str) -> Option<&T> {
+        if let Some(resources) = self.inner.get(cap) {
```
danbugs, line 50: I remember you had expressed interest in having capabilities of different resources have the same name. I think this still would not fulfill that, right?

crates/slightfile/src/capability_store.rs
```diff
@@ -0,0 +1,177 @@
...
+                // check if there is an Any resource
+                resources.get(&ResourceName::Any)
+            }
+        } else {
+            None
+        }
+    }
+
+    /// Insert a new capability into the store.
+    pub fn insert(&mut self, name: ResourceName, cap: &str, value: T) -> Option<T> {
```
danbugs, line 63: Once again, change the parameters names.

crates/slightfile/src/capability_store.rs
```diff
@@ -0,0 +1,177 @@
...
+        store.insert(ResourceName::Any, "messaging", state_2.clone());
+        store.insert(
+            ResourceName::Specific("my-other-container".to_owned()),
+            "keyvalue",
+            state_3.clone(),
+        );
+
+        assert_eq!(store.get("my-container", "keyvalue"), Some(&state_1));
+        assert_eq!(store.get("my-pubsub", "messaging"), Some(&state_2));
+        assert_eq!(store.get("my-other-container", "keyvalue"), Some(&state_3));
```
danbugs, line 174: Your examples are not quite aligned. In line 29, you had `store.get(.)` take 1 parameter – a specific resource. Here, you're passing the name `first` (i.e., not `resource`), and a non-specific capability `resource`.

Also, in line 51, you used `name` in `get` as the value to finding a `ResourceName::Specific`. How's that if `my-container` is `name`?

In other examples (i.e., line 29), you used values like `keyvalue.redis` as the value to `ResourceName::Specific`.

danbugs (changes requested):

crates/slightfile/src/capability_store.rs
Mossaka: The comments are old. I will fix that!

crates/slightfile/src/capability_store.rs
Mossaka: Thanks for pointing it out! Line 29 is out-of-date and I forgot to update it.

Also, in line 51, you used name in get as the value to finding a ResourceName::Specific. How's that if my-container is name?

I am not quiet sure if I understand the question. my-container is passed as name parameter to get function.

crates/slightfile/src/capability_store.rs
Mossaka: Umm it's actually not clear to me what names are the best to choose from for thesse variables. Let me try to list different kinds of "names" we have from the slightfile

1. "my-container" is the name for a resource that apps want to refer to when they do `keyvalue::open(.)`
2. "keyvalue" is a name for a capability
3. "keyvalue.redis" is a name for resource

Perhaps we should call 1. as `ResourceName`, 2. as `CapabilityKind` and 3. as `ResourceKind`?

danbugs: How about:
"my-container") "CapabilityGuestName"
"keyvalue") "CapabilityType"
"keyvalue.redis") "CapabilityImplementor"
... ?

crates/slightfile/src/capability_store.rs
danbugs: > I am not quiet sure if I understand the question. my-container is passed as name parameter to get function.

My confusion is w/ the fact that you're using name in line 51 w/ conjunction to `ResourceName::Specific(.)` (e.g., `ResourceName::Specific("my-container")`; while, in other examples (i.e., line 26), you used `ResourceName::Specific(.)` like `ResourceName::Specific("keyvalue.redis")`, which are two different things, but I'm guessing that's just 'cause the doc comments are old.

crates/slightfile/src/capability_store.rs
Mossaka: I like `CapabilityType` more than `CapabilityKind` and `capabilityimplementor` more than `resourcekind`. 

I don't like to mention `Guest` in the slightfile crate as it is designed from the guests' perspectives. I think `resource` is a good word for describing something apps want to get when doing `keyvalue.open` or `messaging::Pub::open()` in a sense that it is acquiring something it owns.

crates/slightfile/src/capability_store.rs
Mossaka: fixed.

crates/slightfile/src/capability_store.rs
Mossaka: I fixed the docs. Does it answer your questions now?

crates/slightfile/src/slightfile.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// For each capability, deduplicate the resource names.
+    ///
+    /// For example, if you have two resources with the same resource name,
+    /// this will return only one resource.
+    ///
+    /// A special case is when you have a resource that uses the
+    /// Any resource name. In this case, all reousrces of the same capability
+    /// except this one will be removed from the list.
+    pub fn de_dup(self) -> Result<Self> {
+        unimplemented!("de-duplicate logic is not implemented yet");
```
danbugs, line 76: 👍

crates/slightfile/tests/good/dup_good-1.toml
```diff
@@ -0,0 +1,12 @@
+specversion = "0.2"
```
danbugs, line 1: ~change this file's name to match the other ones using hyphens rather than underscores.

danbugs (commented):
This is mostly lgtm~ I'll be testing it locally now and comment if I run into any issues. If not, I'll just approve and I'll be ok w/ merging after we settle on the name change.

danbugs:
I've just tested this locally, and it worked as I expected it – good stuff, @Mossaka . You've also done a great job on test coverage 👍 

Again, I'm good to merge this once we settle on naming, and if you don't mind about not having de_dup implemented rn.

danbugs (approved):

crates/slightfile/src/capability_store.rs
Mossaka: it will fulfill that

PR #384, opened 2023-04-15
chore: bump to 0.5.0

Mossaka:


danbugs (approved):
LGTM~ Let's chat after KubeCon about triggering a release. We could prob. even include JS guest apps support by then.

PR #385, opened 2023-04-15
fix: change receive to receive with a timeout

Mossaka:
This PR serves as a temperary solution to unblock receive method. We can iterate through it to provide a better interface for `receive-timeout` method.


danbugs (approved):
LGTM

PR #388, opened 2023-04-17
fix(messaging): fixed a bug in `subscribe` function and moved filesystem implementation in house

Mossaka:
Since the filesystem implementation is rather simply, it is better to be implemented entirely within the same crate as this will reduce the amount of complexity to fixing some issues or adding new features. 

This PR moves the filesystem implementation from a git repo to the messaging crate. It also fixes a bug in the `subscribe` function where it copies the file without checking if the `from` file exists. 

These few lines are newly added

```rust
// If the topic file doesn't exist, create it
if !topic_path.exists() {
    fs::File::create(&topic_path)?;
}
```

This PR also implements a destroyable temp directory so that each time the local direction drops, the directories being created will also drop.


danbugs:
I think there's value to having the implementation not be in house – this way, we can maintain SpiderLightning issues being more targeted towards the runtime wrapping itself rather than focusing on specific crate implementations.

Mossaka:
> I think there's value to having the implementation not be in house – this way, we can maintain SpiderLightning issues being more targeted towards the runtime wrapping itself rather than focusing on specific crate implementations.

For service implementations, I totally agree. Do you have a plan to publish the crate to "crates.io"? I want to reduce the number of deps on repos that just live on Github. There are some downsides to it:
1. we can't tell if there is a breaking change or not based on the git commit rev
2. it's also difficult to audit deps. Admitted we are not doing security audit but this might be something we wanna do in the future. 

I am fine with in-house implementation of all the local services for filesystems as they are relatively straight forward and usually will be the first crate to be updated when we experiment different interface APIs. This allows us to quickly iterate through it.

danbugs:
I'd be okay w/ publishing the crate, but, if this is the route you prefer to go to, I don't have a strong preference. I'll approve – up to you!

danbugs (approved):

PR #389, opened 2023-04-17
fixed typo in slightfile of nats messaging example

danbugs:


Mossaka (approved):
Not sure if this is a typo, because it displays that the variable name ("NATS_CREDS") doesn't have to be the same as the secret's name ("NATS_CREDS_PATH").

danbugs:
It is – our Azure credential is called "NATS_CREDS_PATH", and not "NATS_CREDS".

PR #392, opened 2023-04-25
CI: Update Windows CI to the correct openssl path

Mossaka:
The current path to openssl.exe on Win agents is `C:\Program Files\OpenSSL`


Mossaka:
Will merge this in and let #393 to fix the clippy issue

PR #394, opened 2023-04-26
fix: remove tempdir crate and replace it with tempfile

Mossaka:
The tempdir crate was an out-of-date crate that depends on a older version of `rand`, which causes some issues on our tests when we import `Alphanumeric` from `rand`. 

Since `tempfile` is a crate meant to replace `tempdir`, this PR replaces it to `tempfile`. 

This should make #384 pass the CI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2006,12 +2006,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


danbugs (approved):
LGTM

PR #395, opened 2023-04-29
merging slightJS upstream

danbugs:
This brings:
- SlightJS's build command, and
- a new JS template (which allows users to make use of SlightJS's engine).

PS Also, I had to update wasmtime to 8.0.1 to match Wizer's version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 reqwest = "0.11"
 flate2 = "1"
 tar = "0.4"
+wizer = "1.6.0"
 
 [dev-dependencies]
 tempfile = { workspace = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 reqwest = "0.11"
 flate2 = "1"
 tar = "0.4"
-wizer = "1.6.0"
+wizer = { git = "https://github.com/bytecodealliance/wizer", rev = "3acc39cc561e7f985f163a2c8e89259a0dfc2f2f"}
 
 [dev-dependencies]
 tempfile = { workspace = true }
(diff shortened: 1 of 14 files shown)


src/commands/build.rs
```diff
@@ -0,0 +1,44 @@
...
+    fs::{self, File},
+    process::Command,
+};
+use wizer::Wizer;
+
+use anyhow::Result;
+
+pub fn handle_build(engine_path: &str, js_path: &str, output_file: &str) -> Result<()> {
+    if env::var("JS_COMPILED").eq(&Ok("1".into())) {
+        env::remove_var("JS_COMPILED");
```
Mossaka, line 12: Probably it's not a good idea to remove user's env var even if its rarely seen in other use cases.

src/cli.rs
```diff
@@ -46,6 +46,16 @@ pub enum Commands {
         #[clap(short, long, value_parser = InterfaceParser)]
         name_at_release: InterfaceAtRelease,
     },
+
+    /// Build a JS Slight project
+    Build {
```
Mossaka, line 51: If this comamnd is JS specific, could we rename it to BuildJS?

templates/js/slightjs_engine.wasm
Mossaka, line 1: Is there anyway to put the source code in templates, instead of a compiled wasm module? We can have the necessary commands in the Makefile to compile that js source code to this wasm module in the CI / CD.

Mossaka (commented):
Did my first round of review. It's a good start! I left a few commands.

templates/js/slightjs_engine.wasm
danbugs: No~ We don't want users building the engine by themselves. The source code is in Rust.

src/cli.rs
danbugs: I figured I'd keep it `build` in case we want to expand to other languages in the future, but I'm good w/ `BuildJs`.

src/commands/build.rs
danbugs: This script is setting the env. var. just before it, so we're just removing what we've set.

Cargo.toml
```diff
@@ -77,10 +78,10 @@ slight-http-client = { path = "./crates/http-client" }
 slight-http-api = { path = "./crates/http-api" }
 wit-bindgen-wasmtime = { git = "https://github.com/fermyon/wit-bindgen-backport", features = ["async"] }
 wit-error-rs = { git = "https://github.com/danbugs/wit-error-rs", rev = "05362f1a4a3a9dc6a1de39195e06d2d5d6491a5e" }
-wasmtime = "7.0.0"
-wasmtime-wasi = "7.0.0"
-wasi-common = "7.0.0"
-wasi-cap-std-sync = "7.0.0"
+wasmtime = "8.0.1"
```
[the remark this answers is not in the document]
danbugs: ~iirc, Fermyon updated.

PR #401, opened 2023-05-12
update the perm mode of install.sh

Iceber:
Add the executable permission to install.sh for local running


danbugs (approved):
LGTM – Thank you for your contribution!

PR #411, opened 2023-07-31
Fix typo in Cargo.toml

mboersma:
A typo in `Cargo.toml` was causing unnecessary warnings in dev builds:

```shell
% cargo build
warning: /Users/matt/projects/spiderlightning/Cargo.toml:
  dependency (rand) specified without providing a local path, Git repository, version, or workspace dependency to use. This will be considered an error in future versions
warning: /Users/matt/projects/spiderlightning/Cargo.toml:
  unused manifest key: dev-dependencies.rand.worspace
...
```

It builds silently after this change.


Mossaka (approved):
LGTM! Thanks

PR #419, opened 2023-08-31
Windows installer - Remove Update-SessionEnvironment dependency

fcabrera23:
The [Update-SessionEnvironment](https://docs.chocolatey.org/en-us/create/functions/update-sessionenvironment) is an implementation of Chocolatey to refresh all the session environment variables. If Chocolatey is not installed, this will fail.  Given that the only variable that is updated is $Path, we only need to refresh that one for the current session, and we can do that manually.


danbugs (approved):
LGTM – Thanks for contributing!

